PR #8, opened 2025-01-20
Add variable for core-web-app docker image url

danifr:


variables-common.tf
```diff
@@ -85,6 +85,12 @@ variable "bbpssh_cidr" {
   sensitive   = false
 }
 
+variable "core_web_app_docker_image_url" {
```
heerener, line 88: Maybe without the `_url` suffix? It's a nitpick though :-)

heerener (approved):

PR #19, opened 2025-01-22
Explicitly define the provider which gets inherited within the module

DriesVerachtert:
https://github.com/openbraininstitute/PMO/issues/50

### DIFF
--- a/nexus/blazegraph/provider.tf
+++ b/nexus/blazegraph/provider.tf
@@ -1,0 +1,8 @@
+terraform {
+  required_providers {
+    aws = {
+      source  = "hashicorp/aws"
+      version = ">= 2.7.0"
+    }
+  }
+}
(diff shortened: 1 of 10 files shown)


danifr:
what is the nexus module doing differently that needs those specific provider terraform block?

DriesVerachtert:
@danifr Within the nexus module, additional providers are defined, each time with a different set of default tags. For example the nexus blazegraph module has a provider with default tags SBO_Billing=nexus and Nexus=blazegraph. The disadvantage is that we now have to explicitly define that each module depends on the AWS provider which gets inherited from the root, but the advantage is that you do not have to define tags anymore in any resource in the nexus modules. Also the tagging works much better: for example the nexus ECS containers are correctly tagged, while that's not the case for other modules like BlueNaas => we'll likely need to apply the same mechanism to other modules to get tagging to work everywhere.

danifr:
Thanks for the explanation, all clear!!

danifr (approved):

PR #22, opened 2025-01-22
Cleanup bluenaas tagging: use default tags and propage tags to ECS tasks/containers

DriesVerachtert:


main.tf
```diff
@@ -192,6 +192,16 @@ module "nse" {
   me_model_analysis_docker_image_url = "bluebrain/me-model-analysis:latest"
 }
 
+provider "aws" {
```
danifr: Is this block meant to be here in the main.tf file?

DriesVerachtert: It defines an AWS provider which is an alias of the main AWS provider but which has the default tags for bluenaas. That's then passed on to the bluenaas module so that everything defined within that module should automatically get that tag.

It's done the same in the nexus module, but there there's a layer in between:
* main.tf passes by default the regular AWS provider to the nexus module
* within the nexus module, a bunch of alias providers are defined, each with a different set of default tags per submodule and each submodule is imported with its own alias provider
* each submodule uses the alias provider so all resources have a set of default tags

danifr: But why not define this block within the bluenaas module? What if you want to do the same for the CS module?
can you add another `provider "aws" {` block with a different alias?

DriesVerachtert: As far as I understand, you cannot define it within the module itself, but have to create such an alias outside of the module and pass it on to the module. That's also how nexus is doing it with their module in between main.tf and the submodules with the actual resource definitions.

You can add several alias providers as long as they have a different alias name: we can definitely still add others for cs for example. You can see an example in the nexus module where a lot of alias providers are defined for each of the submodules: https://github.com/openbraininstitute/aws-terraform-deployment/blob/staging/nexus/provider.tf

DriesVerachtert: @danifr I wrote before that you have to declare the provider alias outside the module, but I was wrong. This syntax seems to define the alias with the default tag within the module.

danifr: now looks great! thanks!

danifr (approved):

PR #32, opened 2025-01-24
Introduce variables to toggle nexus on and off

heerener:


nexus/variables.tf
```diff
@@ -78,3 +78,15 @@ variable "is_production" {
 variable "nexus_secrets_arn" {
   type = string
 }
+
+variable "is_nexus_openscience_running" {
+  description = "Whether OpenScience nexus should be running"
+  type        = bool
+  default     = true
```
danifr: I would remove default values, IMO it is better if the plan/apply fails before even starting than an unwanted outcome because of a missing variable in the `<enviroment>.tfvars` file.

heerener: Good point - that would have revealed issues I had in development a lot earlier too 😭

danifr (approved):
🥳

DriesVerachtert (approved):

heerener:
I'm taking a different approach which involves a refactor of the nexus code; this PR is now obsolete.

PR #41, opened 2025-01-31
Add backup config for Keycloak DB

danifr:


DriesVerachtert (approved):
It says "module.cs.module.keycloak.aws_db_instance.keycloak_database will be updated in-place" so I guess it's OK :-)

danifr:
there is only one way to know :) 

I just took an snapshot of the staging DB just in case

PR #43, opened 2025-01-31
Add accounting base URL to vlab deployment

pgetta:
Add accounting service base URL to Virtual Lab API to enable accounting integration.

FYI, @bilalesi

### DIFF
--- a/main.tf
+++ b/main.tf
@@ -270,7 +270,7 @@ module "accounting_svc" {
   dockerhub_credentials_arn       = local.dockerhub_bbpbuildbot_secret_arn
   dockerhub_access_iam_policy_arn = local.dockerhub_bbpbuildbot_policy_arn
 
-  root_path = "/api/accounting"
+  root_path = var.accounting_base_path
 }
 
 module "kg_inference_api" {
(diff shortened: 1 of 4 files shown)


pgetta:
Btw, the change will have to be replicated for production env as well. What's the process to do so?

danifr:
After the change is verified in staging we can deploy in production too

heerener (approved):

danifr:
@pgetta the idea is to deploy this change (alongside others) in prod today, can you please verify that works as intended in https://staging.openbluebrain.com/app ? thanks

pgetta:
@danifr Not everything is deployed to staging (not easy as we don't have CI/CD for projects at this point). But, I've already done this change in production manually. Right now my goal is to deploy to prod, so next time TF apply is executed - it's not going to revert my manual changes.

If something breaks - I should be able to apply some bandaid.

danifr:
OK. we just deployed the change in prod, https://github.com/openbraininstitute/aws-terraform-deployment/actions/runs/13071860081/job/36475056626

thanks for your answer :)

pgetta:
Thank you @danifr

PR #47, opened 2025-02-03
Set correct Nexus Delta URL for BlueNaas service

pgetta:
Default value in BlueNaas repo is incorrectly pointing to the production one (which should be fixed separately).
Now it's going to use staging nexus instead.


danifr (approved):
Great thank you!

PR #50, opened 2025-02-04
Add accounting URL to BlueNaas service

pgetta:
The change enables BlueNaaS service integration with accounting.


danifr (approved):

pgetta:
Thank you, @danifr

PR #58, opened 2025-02-05
Enable AWS cost optimization hub for the whole organization.

jplanasc:


main.tf
```diff
@@ -278,6 +278,11 @@ module "accounting_svc" {
   root_path = var.accounting_base_path
 }
 
+module "billing_cost_management" {
+  source     = "./billing_cost_management"
+  aws_region = local.aws_region
```
danifr: Since there is no `aws_region` module variable, you need to remove this line

jplanasc: Thanks!

danifr:
@jplanasc not 100% sure but I think we should add some logic to your code so it only tries to enable it in prod.

On staging and any other accounts different than it won't work.
Screenshot 2025-02-06 at 09 31 10


you can add something like https://github.com/openbraininstitute/aws-terraform-deployment/blob/staging/nexus/nexus-openscience.tf#L7

DriesVerachtert (dismissed):

jplanasc:
@danifr good point about the production environment. I'll try to add, but not sure the resource will accept `count`, let's see.
Thanks!

jplanasc:
@danifr I think it should be ok now. Do you have further comments?
Thanks!

billing_cost_management/variables.tf
```diff
@@ -0,0 +1,7 @@
+variable "is_production" {
+  type        = bool
+  default     = true
```
danifr: Maybe better to remove the default value, so it fails directly if the variable is not set up

danifr (approved):

PR #63, opened 2025-02-06
Use a different (staging) tag for VLab API in staging

pgetta:
This a temporary solution, just not to touch the `:latest` image used in production.

Once we set up container registry and CI, staging should use `:latest`, and production - version tags.


danifr (approved):
👍

pgetta:
Thanks!

PR #65, opened 2025-02-06
Fix keycloak health check

danifr:
After the upgrade to v25 the /health endpoint runs on the management port


danifr:
Relates to https://github.com/openbraininstitute/INFRA/issues/42

DriesVerachtert (approved):

PR #69, opened 2025-02-10
Make Keycloak ECS task size configurable

danifr:
https://github.com/openbraininstitute/INFRA/issues/53

### DIFF
--- a/cs/keycloak/variables.tf
+++ b/cs/keycloak/variables.tf
@@ -51,3 +51,12 @@ variable "keycloak_secrets_arn" {
   description = "ARN of the Keycloak secrets manager"
   sensitive   = false
 }
+
+variable "keycloak_task_size" {
+  type = object({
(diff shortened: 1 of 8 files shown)


cs/keycloak/variables.tf
```diff
@@ -51,3 +51,12 @@ variable "keycloak_secrets_arn" {
   description = "ARN of the Keycloak secrets manager"
   sensitive   = false
 }
+
+variable "keycloak_task_size" {
+  type = object({
+    cpu    = any
```
heerener: shouldn't these be `number`s? (same for the higher levels of variable declarations)

danifr: Yeah you are probably right. I will change it :)

heerener (dismissed):

heerener (approved):

PR #78, opened 2025-02-13
Add deployment env to core-web-app

bilalesi:
this variable allow the frontend to enable some features depends on the env or tag specific data based on the deployment
ex: for newsletter we need to tag the subscriber in staging to deal with test users (delete)

### DIFF
--- a/variables-common.tf
+++ b/variables-common.tf
@@ -62,6 +62,12 @@ variable "core_web_app_docker_image_url" {
   description = "docker image for the core-web-app"
   sensitive   = false
 }
+variable "core_web_app_deployment_env" {
+  default     = "production"
+  type        = string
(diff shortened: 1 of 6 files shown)


pgetta (dismissed):
LGTM

DriesVerachtert (dismissed):

variables-common.tf
```diff
@@ -62,6 +62,12 @@ variable "core_web_app_docker_image_url" {
   description = "docker image for the core-web-app"
   sensitive   = false
 }
+variable "core_web_app_deployment_env" {
+  default     = "production"
```
danifr, line 66: I would remove the default value

bilalesi: done

pgetta (approved):

PR #110, opened 2025-02-19
Provide accounting base URL for the core web app

pgetta:


danifr:
It took me one or two minutes to figure out where the value of `var.accounting_base_path` was coming from.

Would it make sense to delete all those variables (with a default value) from the variables-common.tf file and simply define them in the locals block of the main.tf file? 

Those are the kind of variables that rarely change.

pgetta:
@danifr yeah, I did question that myself. Let me do that in a separate MR if you are fine with that.

danifr (approved):
Cool, we can do it another PR

pgetta:
Thank you!

PR #117, opened 2025-02-21
Serve core web app from /

pgetta:
The ALB rule path pattern has been removed so that "by default" the requests not matched by the higher priority rules will be handed to core web app, so that it can serve content for the `/`.


danifr (approved):

pgetta:
Much appreciated

PR #138, opened 2025-02-27
Make accounting process oneshot job charges more frequently

pgetta:
The charging cycle has a delay between iterations defaulting to 600 s, which is too slow.
From the user prospective it's best to process job charges as soon as possible, so that job reports appear almost in real-time in the user dashboard.

On the other hand we need to manage the pressure on the db. With the load in coming months I think we should be good with setting that delay at 30 s, otherwise we can adjust and further look into how to optimise things.

FYI, @GianlucaFicarelli


danifr (approved):
ok for me but I am no expert :)

DriesVerachtert (approved):

PR #142, opened 2025-02-27
add elastic cache to virtual lab service

bilalesi:
adding elastic cache functionality to the virtual lab manager.

fyi: @pgetta

### DIFF
--- a/virtual-lab-manager/container.tf
+++ b/virtual-lab-manager/container.tf
@@ -53,9 +53,7 @@ resource "aws_vpc_security_group_ingress_rule" "virtual_lab_manager_allow_port_8
   cidr_ipv4   = var.vpc_cidr_block
   description = "Allow port 8000 http"
 
-  tags = {
-    SBO_Billing = "virtual_lab_manager"
-  }
+  tags = var.virtual_lab_manager_tags
 }
 
 resource "aws_vpc_security_group_egress_rule" "virtual_lab_manager_allow_outgoing_tcp" {
@@ -67,9 +65,7 @@ resource "aws_vpc_security_group_egress_rule" "virtual_lab_manager_allow_outgoin
   cidr_ipv4   = "0.0.0.0/0"
   description = "Allow all TCP"
 
-  tags = {
-    SBO_Billing = "virtual_lab_manager"
-  }
+  tags = var.virtual_lab_manager_tags
 }
 
 resource "aws_vpc_security_group_egress_rule" "virtual_lab_manager_allow_outgoing_udp" {
@@ -81,9 +77,7 @@ resource "aws_vpc_security_group_egress_rule" "virtual_lab_manager_allow_outgoin
   cidr_ipv4   = "0.0.0.0/0"
   description = "Allow all UDP"
 
-  tags = {
-    SBO_Billing = "virtual_lab_manager"
-  }
+  tags = var.virtual_lab_manager_tags
 }
 
 resource "aws_ecs_task_definition" "virtual_lab_manager_ecs_definition" {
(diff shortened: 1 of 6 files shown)


virtual-lab-manager/elasticache.tf
```diff
@@ -0,0 +1,37 @@
...
+    protocol    = "tcp"
+    cidr_blocks = [var.vpc_cidr_block]
+    description = "allow access from within VPC"
+  }
+  tags = var.virtual_lab_manager_tags
+}
+
+resource "aws_elasticache_subnet_group" "vlm_redis_subnet_group" {
+  name       = "vlm-redis-subnet-group"
+  subnet_ids = [aws_subnet.virtual_lab_manager_a.id, aws_subnet.virtual_lab_manager_b.id]
```
pgetta, line 22: Over time it's best to create a separate subnet for this so that security policy for each can be adjusted accordingly to the service needs.

virtual-lab-manager/elasticache.tf
```diff
@@ -0,0 +1,37 @@
...
+  name        = "vlm-redis-sg"
+  description = "Security group for Redis instance for Virtual Lab Manager"
+
+  vpc_id = var.vpc_id
+
+  ingress {
+    from_port   = 6379
+    to_port     = 6379
+    protocol    = "tcp"
+    cidr_blocks = [var.vpc_cidr_block]
```
pgetta, line 14: The subnet itself is a private one, meaning there is already no access from outside of VPC.
I would recommend to put a wildcard here for now, so there is no confusion: removing the value here doesn't make Redis publicly accessible.

pgetta (approved):
Looks good Bilal,

I've left couple of comments, one for the long term and another maybe worth to address rather sooner if you think it makes sense.

virtual-lab-manager/elasticache.tf
bilalesi: yeah sorry was not what we discussed, i want to just make things work and then improve it, i will do it as 2nd step

virtual-lab-manager/elasticache.tf
bilalesi: do you mean ports to be 0-0 ? and change the protocol ?

pgetta: For the ports, the current values are actually good security wise.

For the cidr_blocks:
```suggestion
    cidr_blocks = ["0.0.0.0/0"]
```

bilalesi:
@openbraininstitute/infrastructure please can you have a look 🙏

DriesVerachtert (approved):

virtual-lab-manager/container.tf
```diff
@@ -216,6 +210,14 @@ resource "aws_ecs_task_definition" "virtual_lab_manager_ecs_definition" {
...
           name  = "MAIL_PASSWORD"
           value = var.virtual_lab_manager_mail_password
+        },
+        {
+          name  = "REDIS_HOST"
+          value = aws_elasticache_cluster.vlm_redis_cluster.cache_nodes[0].address
+        },
+        {
+          name  = "REDIS_PORT"
+          value = aws_elasticache_cluster.vlm_redis_cluster.port
```
DriesVerachtert, line 220: Hi @bilalesi , as you probably saw: the deployment of the container failed. Not 100% sure but I think it complains about the name&value environment variables which all have to be strings, and likely this port is some integer.
You could replace that value line with for example:
```
value = "${aws_elasticache_cluster.vlm_redis_cluster.port}"
```

bilalesi:
sorry @DriesVerachtert , i was the whole day in meeting, i did not saw the message, i will fix tomrrow

danifr:
@bilalesi no worries it was fixed by Erik here https://github.com/openbraininstitute/aws-terraform-deployment/pull/149/files#diff-12ffbe731120ee9910c3f8908d15b6d23067c7f325768717edd332e105b14c91R220

PR #147, opened 2025-03-03
Temporarily allow HTTP 307 on the core web app health check endpoint

pgetta:
The domain redirect is now enabled on the core web app server itself.

When ALB sends a health check the hostname in the request never matches the "primary" domain so web app responds with redirect. This is a temporary workaround. To be reverted back when domain redirect either is disabled or implemented on AWS ALB level.


danifr (approved):

pgetta:
Thank you Daniel

pgetta:
@danifr  I've got a green light to deploy the fix to enable domain redirects (within the core web app), would it be possible to deploy this change to production?

danifr:
Yes, on it https://github.com/openbraininstitute/aws-terraform-deployment/actions/runs/13630479660/job/38096968446 :)

pgetta:
Thanks!

PR #156, opened 2025-03-13
Install tljh v0.2.0 (jupyterlab v3) on Ubuntu 22.04 and other fixes

danifr:
WebIO is not compatible with jupyterlab v4 so we need an older version. 
Newer versions of Ubuntu come with a newer python version that it is not compatible with tljh v0.2.0


danifr:
workaround for: https://github.com/openbraininstitute/prod-launch-notebook/issues/16

DriesVerachtert (approved):

PR #160, opened 2025-03-14
Neuroagent new features

jankrepl:
**TODO**
- [x] Check the current deployment status
   - [x] https://www.staging.openbraininstitute.org/api/agent - not running (no logs even)
   - [x] https://www.openbraininstitute.org/api/agent - not running (no logs even)
- [x] Create bucket (name should be coming from the `staging.tfvars/production.tfvars`
- [x] Add policies to ECS IAM role to read/write on the bucket
- [x] Make sure if that if frontend url not provided the LLM does not try to propose a viewer url (that logic is inside the code already)
- [x] (neuroagent) Make sure that access/secret key not provided (or null) will lead to aws reading the profile
- [x] (neuroagent) Build and push new tag 
- [x] Update ECS environment variables
- [x] Put tavily API key in the secrets manager (prod/staging)
- [x] Bump up neuroagent image in terraform to `v0.4.1`

### DIFF
--- a/ml/agent.tf
+++ b/ml/agent.tf
@@ -60,32 +60,8 @@ module "ecs_service_agent" {
           value = "https://${var.nexus_domain_name}/api/nexus/v1"
         },
         {
-          name  = "NEUROAGENT_AGENT__MODEL"
-          value = "simple"
-        },
-        {
           name  = "NEUROAGENT_OPENAI__MODEL"
           value = "gpt-4o-mini"
-        },
-        {
-          name  = "NEUROAGENT_TOOLS__MORPHO__SEARCH_SIZE"
-          value = "10"
-        },
-        {
-          name  = "NEUROAGENT_TOOLS__KG_MORPHO__SEARCH_SIZE"
-          value = "6"
-        },
-        {
-          name  = "NEUROAGENT_TOOLS__TRACE__SEARCH_SIZE"
-          value = "10"
-        },
-        {
-          name  = "NEUROAGENT_TOOLS__LITERATURE__RERANKER_K"
-          value = "8"
-        },
-        {
-          name  = "NEUROAGENT_KEYCLOAK__VALIDATE_TOKEN"
-          value = "true"
         },
         {
           name  = "NEUROAGENT_KEYCLOAK__ISSUER"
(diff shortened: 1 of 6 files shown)


ml/agent.tf
```diff
@@ -38,72 +52,32 @@ module "ecs_service_agent" {
...
-          name  = "NEUROAGENT_TOOLS__TRACE__SEARCH_SIZE"
-          value = "10"
+          name  = "NEUROAGENT_DB__PORT"
+          value = module.ml_rds_postgres.db_instance_port
         },
         {
-          name  = "NEUROAGENT_TOOLS__LITERATURE__RERANKER_K"
-          value = "8"
+          name  = "NEUROAGENT_DB__PREFIX"
+          value = "postgresql://"
```
WonderPG: Should contain the async plugin i.e. `postgresql+asyncpg://`

ml/agent.tf
```diff
@@ -114,26 +88,30 @@ module "ecs_service_agent" {
           value = "https://openbrainplatform.org, https://bbp.epfl.ch"
```
WonderPG: epfl could be removed

ml/agent.tf
```diff
@@ -38,72 +52,32 @@ module "ecs_service_agent" {
...
-          name  = "NEUROAGENT_DB__PORT"
-          value = module.ml_rds_postgres.db_instance_port
+          name  = "NEUROAGENT_KNOWLEDGE_GRAPH__BASE_URL"
+          value = "https://${var.nexus_domain_name}/api/nexus/v1"
         },
         {
-          name  = "NEUROAGENT_DB__USER"
-          value = module.ml_rds_postgres.db_instance_username
+          name  = "NEUROAGENT_KNOWLEDGE_GRAPH__DOWNLOAD_HIERARCHY"
+          value = "true"
```
WonderPG: iirc we removed this env var

danifr:
TF wise looks good. One thing you can fix for the future is https://github.com/openbraininstitute/aws-terraform-deployment/blob/9142549e50fd718c0897522b5daae4a197d79b78/ml/agent.tf#L84 you can use `${var.primary_domain}`

unless you want those values for another reason :)

WonderPG (commented):
Thank you !

ml/agent.tf
jankrepl: Done in 6972ae8

ml/agent.tf
jankrepl: Done in 6972ae8

WonderPG (dismissed):

WonderPG (dismissed):

WonderPG (approved):

ml/agent.tf
```diff
@@ -81,7 +81,7 @@ module "ecs_service_agent" {
         },
         {
           name  = "NEUROAGENT_MISC__CORS_ORIGINS"
-          value = "https://openbraininstitute.org,https://www.openbraininstitute.org,https://staging.openbraininstitute.org,https://www.staging.openbraininstitute.org"
+          value = "https://${var.primary_domain},https://www.${var.primary_domain}"
```
danifr, line 84: Not a big deal but ${var.primary_domain}'s value is `www.openbraininstitute.org` so no need to add the www there.
Just `https://${var.primary_domain}` will work

PR #162, opened 2025-03-17
add discount env vars to vlab

bilalesi:


virtual-lab-manager/container.tf
```diff
@@ -222,6 +222,18 @@ resource "aws_ecs_task_definition" "virtual_lab_manager_ecs_definition" {
         {
           name  = "STRIPE_DEVICE_NAME"
           value = "Open Brain Institute"
+        },
+        {
+          name  = "ENABLE_DISCOUNT"
+          value = true
```
heerener:
```suggestion
          value = "true"
```

See also line 116; you might have to modify the applacation to make sure it can deal with this being a string instead of a bool.

bilalesi: fixed, thanks @heerener

danifr (approved):

PR #168, opened 2025-03-18
Add a separate instance of Core Web App for entitycore migration

pgetta:
I've also taken a chance to refactor the module a bit: made naming consistent, removed redundant module params, outdated prefixes etc. I hope I didn't forget anything.

P.S. Also looked at updating core web app to support custom base path but that would be way more complex and error prone.


pgetta:
Still need to update CI for core web app to build the image, doing that now.

pgetta:
Ok, CI fails, let me fix it quickly.

pgetta:
Ok, that's better. CI passes now.

pgetta:
I've added a hostname `next.staging.openbraininstitute.org` but maybe you have a better name in mind.

core_webapp/variables.tf
```diff
@@ -34,32 +52,28 @@ variable "vpc_cidr_block" {
...
 
 variable "allowed_source_ip_cidr_blocks" {
-  type = list(string)
-}
-
-variable "core_webapp_secrets_arn" {
-  type        = string
-  description = "coreweb-app secrets ARN"
-  sensitive   = false
+  default     = ["0.0.0.0/0"]
```
danifr: Not a big fan of having default values that are so open

pgetta: This is not actually used at the moment as all web apps are publicly accessible. I kept it for later, if/when, for example, we want to limit staging access to our internal OBI network/vpn.

I've made a change to set a default value to null to avoid the creation of corresponding condition in the ALB listener rule, but apparently AWS requires at least one condition to be present, so I put that value in the default to be less verbose in the main.tf.

Do you think we should have this explicitly in main.tf instead of a default? (still, the value will be `0.0.0.0/0` for all core web app deployments for the time being).

danifr: IMO it is better if we have it in main.tf instead of "hidden" as a default value.
I had some bad experiences with default values in the past hehe

pgetta: Done, I've moved those to the `main.tf`

danifr (approved):
thanks Pavlo!

pgetta:
Thank you, @danifr .

Before merging, let me check couple of things on the web app repo side.

pgetta:
Ok, the image was built.

I guess I'll try to merge and see if the TF apply and deployment work.

And whenever you have time to add domain records we should have routing working.

PR #175, opened 2025-03-19
Enable on-demand svc from outside of EPFL network

pgetta:


on_demand_svc/src/ws_authz.py
```diff
@@ -1,19 +1,15 @@
...
 
     # FIXME add proper keycloak authz
-    if ((ip in epfl_cidr or ip in bbp_dmz_cidr) and token.startswith("Bearer ")):
+    if token.startswith("Bearer "):
         authResponse = generateAllow("me", event["methodArn"])
         # FIXME get user vlab/proj
-        authResponse["context"] = {"SVC_VLAB": "O1_data_physiology_sep11",
-                                   "TOKEN": token}
+        authResponse["context"] = {
+            "SVC_VLAB": "O1_data_physiology_sep11",
```
heerener: Should this be a repository secret (and changed to a different value), since this repo is public? Or doesn't it matter that anyone can use this info?

pgetta: Not an issue, as it's not even used under the hood.
There is quite some work to be done with this service, many things are not fully implemented.

I've updated it to be `dummy-vlab-id` so it's clear what it is.

heerener (approved):

pgetta:
Thank you, Erik!

PR #180, opened 2025-03-20
Fix entitycore_svc apply issues

eleftherioszisis:
* Use unique name for staging s3 bucket

### DIFF
--- a/sandbox-nse.tfvars
+++ b/sandbox-nse.tfvars
@@ -38,5 +38,5 @@ hpc_resource_provisioner_sbo_nexusdata_bucket = "s3://sbonexusdata-sandbox"
 hpc_resource_provisioner_containers_bucket    = "s3://sboinfrastructureassets-sandbox/containers"
 hpc_resource_provisioner_scratch_bucket       = "s3://sbosandbox-lec3cn"
 
-entitycore_svc_s3_bucket_name = "entitycore-data-staging"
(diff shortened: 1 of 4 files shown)


sandbox-nse.tfvars
```diff
@@ -38,5 +38,5 @@ hpc_resource_provisioner_sbo_nexusdata_bucket = "s3://sbonexusdata-sandbox"
 hpc_resource_provisioner_containers_bucket    = "s3://sboinfrastructureassets-sandbox/containers"
 hpc_resource_provisioner_scratch_bucket       = "s3://sbosandbox-lec3cn"
 
-entitycore_svc_s3_bucket_name = "entitycore-data-sandbox-nse-test"
+entitycore_svc_s3_bucket_name = "entitycore-data-staging"
```
danifr: We probably want theses 2 in `production.tfvars` file:
```
entitycore_svc_s3_bucket_name = "entitycore-data-production"
entitycore_svc_image_url      = "public.ecr.aws/openbraininstitute/entitycore:2025.3.0"
```
Otherwise when de deploy in prod things will fail.

GianlucaFicarelli (approved):

sandbox-nse.tfvars
eleftherioszisis: Indeed. Added.

PR #207, opened 2025-03-26
Update VLab API welcome bonus related env vars

pgetta:
Rename + set amount of credits from 200 to 100 as per the latest requirement.


danifr (approved):

pgetta:
Thanks!

PR #213, opened 2025-03-27
Enable Sentry for the core web app

pgetta:
The secret needs to be updated. Please ping me whoever will be looking at this ticket, so I can send the value.


danifr:
@pgetta I can take care of it

danifr (approved):

PR #214, opened 2025-03-27
Enable Sentry for the vlab api

pgetta:
I've added a separate `var.deployment_env` variable for convenience, to avoid `var.is_staging ? "staging" : "production"` switches in multiple places.

SENTRY_DSN in corresponding secret has to be updated for both production and staging. Please ping me so I can provide the value.


danifr (dismissed):

pgetta:
To be merged and tested after the release.

PR #216, opened 2025-03-28
public release 2025.03.27.1 for corewebapp and virtual-lab-svc

bilalesi:


DriesVerachtert (approved):
let's deploy to prod :-)

PR #224, opened 2025-04-01
On-demand svc: create capacity provider only for EC2 tasks

pgetta:
Because for FARGATE instances this blocks provisioning.


danifr (approved):
I guess you are aware but the change will destroy the capacity provider in staging :)

pgetta:
Thanks !

pgetta:
> I guess you are aware but the change will destroy the capacity provider in staging :)

Thanks for checking and yes, that is expected.

PR #226, opened 2025-04-01
Migrate me-model-analysis image to AWS ECR

pgetta:


danifr (approved):
LGTM!

pgetta:
@danifr could you please check the deployment, it seems it failed because TF couldn't delete some iAM policy. Maybe that should be done by hand to address the issue?

danifr:
Checking

danifr:
I manually detached the policy and then delete it.

Not sure why terraform does not do it automatically. I will probably need to do the same when we deploy in prod.

Not a big deal in any case.

PR #228, opened 2025-04-02
Add permission for lambda connect to pass roles to ECS task

pgetta:
Why it works without that permission when using ECS on EC2 is still to be investigated.

Separate thank you to @danifr to help debug the issue.


danifr (dismissed):
🫡

danifr (approved):

PR #231, opened 2025-04-02
Pass deployment env name to me-model-analysis

pgetta:


pgetta:
I've added a separate `var.deployment_env` variable for convenience, to avoid `var.is_staging ? "staging" : "production"` switches in multiple places.

This is going to be used later on other places, e.g. Sentry integration: https://github.com/openbraininstitute/aws-terraform-deployment/pull/214, also for other services.

DriesVerachtert (approved):

pgetta:
Thank you Dries

PR #234, opened 2025-04-03
clean docker for thumbnail api 

bilalesi:
1. clean docker credentails for thumbnail api 
2. add entitycore uri env var

### DIFF
--- a/thumbnail-generation-api/thumbnail-generation-api-container.tf
+++ b/thumbnail-generation-api/thumbnail-generation-api-container.tf
@@ -85,11 +85,6 @@ resource "aws_iam_role_policy_attachment" "ecs_task_execution_role_attachment" {
 }
 
 
-resource "aws_iam_role_policy_attachment" "thumbnail_generation_api_ecs_task_role_dockerhub_policy_attachment" {
-  role       = aws_iam_role.thumbnail_generation_api_ecs_task_execution_role.name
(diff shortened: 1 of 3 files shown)


thumbnail-generation-api/thumbnail-generation-api-container.tf
```diff
@@ -182,6 +174,10 @@ resource "aws_ecs_task_definition" "thumbnail_generation_api_task_definition" {
           {
             name  = "SENTRY_DSN"
             value = "https://df67a83aab2f208467f94df08a207e59@o224246.ingest.us.sentry.io/4507457507885056"
+          },
+          {
+            name : "ENTITY_CORE_URI",
+            value : "https://staging.openbraininstitute.org/api/entitycore"
```
danifr, line 180: Better if you use something like `https://${var.primary_domain}/api/entitycore` (you might need to add the variable to the module)

bilalesi: it was in purpose, we don't have entity core in prod for the moment, do you think it's okey to keep as is for the moment ?

danifr: Alright then :) just don't forget to change it in the future

danifr (approved):

PR #255, opened 2025-04-09
Bump production core web to v2025.04.09.1

pgetta:
Jean-Denis asked to deploy this to prod right away as it contains an important bug fix.


pgetta:
I presume we don't want to deploy the rest of the changes, so maybe we can just cherry pick this commit so that the rest of the prod stays intact.

danifr (approved):

PR #264, opened 2025-04-15
Add obi-one and obi-generative-gui

GianlucaFicarelli:


obi_generative_gui/backend.tf
```diff
@@ -0,0 +1,259 @@
...
+  description = "Allow port 8000 http"
+}
+
+resource "aws_vpc_security_group_ingress_rule" "obi_generative_gui_allow_in_tcp" {
+  security_group_id = aws_security_group.obi_generative_gui_ecs_task.id
+  # TODO limit to what is needed
+  ip_protocol = "tcp"
+  from_port   = 0
+  to_port     = 65535
+  cidr_ipv4   = "0.0.0.0/0"
```
danifr: we should be more strict with the ingress rules

obi_generative_gui/backend.tf
```diff
@@ -0,0 +1,259 @@
...
+      portMappings = [
+        {
+          hostPort      = 8000
+          containerPort = 8000
+          protocol      = "tcp"
+        }
+      ]
+
+      healthcheck = {
+        command     = ["CMD-SHELL", "exit 0"] // TODO: add a proper health check.
```
danifr: Lets add a proper healthcheck or just simply delete this block

obi_one/backend.tf
```diff
@@ -0,0 +1,259 @@
...
+  description = "Allow port 8000 http"
+}
+
+resource "aws_vpc_security_group_ingress_rule" "obi_one_allow_in_tcp" {
+  security_group_id = aws_security_group.obi_one_ecs_task.id
+  # TODO limit to what is needed
+  ip_protocol = "tcp"
+  from_port   = 0
+  to_port     = 65535
+  cidr_ipv4   = "0.0.0.0/0"
```
danifr: Same as before, this is too open for ingress

obi_generative_gui/alb.tf
```diff
@@ -0,0 +1,40 @@
+resource "aws_lb_target_group" "obi_generative_gui_private_tg" {
+  #ts:skip=AC_AWS_0492
+  name        = "obi-generative-gui-private"
+  port        = 8000
```
danifr: I saw this 8000 port used in a few places, maybe worth to set it as an internal module variable?

obi_one/alb.tf
```diff
@@ -0,0 +1,40 @@
...
+    target_group_arn = aws_lb_target_group.obi_one_private_tg.arn
+  }
+
+  condition {
+    path_pattern {
+      values = ["${var.root_path}*"]
+    }
+  }
+
+  condition {
```
danifr: If this service is meant to be accessible from 0.0.0.0/0 since inception, wouldn't make sense to delete all this `allowed_source_ip_cidr_blocks ` logic? 

I think it made sense when we were at EPFL/BBP but not sure it is useful anymore

danifr (changes requested):
I added some comments :)

obi_one/alb.tf
```diff
@@ -0,0 +1,34 @@
...
+  health_check {
+    enabled  = true
+    path     = "${var.root_path}/health"
+    protocol = "HTTP"
+  }
+}
+
+resource "aws_lb_listener_rule" "obi_one_private_listener_rule" {
+  listener_arn = var.private_alb_listener_arn
+  priority     = 710
```
danifr, line 22: You have set priority 710 for the private_alb_listener_arn for both obi-one and gui. Please change one of them to 711 for example.

obi_generative_gui/backend.tf
```diff
@@ -0,0 +1,241 @@
...
+    Name = "obi_generative_gui"
+  }
+
+  setting {
+    name  = "containerInsights"
+    value = "disabled" #tfsec:ignore:aws-ecs-enable-container-insight
+  }
+}
+
+# TODO make more strict
```
danifr: This is probably not needed anymore, at least not here

obi_generative_gui/backend.tf
```diff
@@ -0,0 +1,241 @@
...
+  security_group_id = aws_security_group.obi_generative_gui_ecs_task.id
+
+  ip_protocol = "tcp"
+  from_port   = var.container_port
+  to_port     = var.container_port
+  cidr_ipv4   = data.aws_vpc.main.cidr_block
+  description = "Allow access to the container port"
+}
+
+resource "aws_vpc_security_group_egress_rule" "obi_generative_gui_allow_outgoing_tcp" {
```
danifr: OPTIONAL 

Line 53 to 71 can be combined into 1 resource:

```
resource "aws_vpc_security_group_egress_rule" "obi_generative_gui_allow_outgoing" {
  security_group_id = aws_security_group.obi_generative_gui_ecs_task.id
  description       = "Allow egress to any destination"
  cidr_ipv4         = "0.0.0.0/0"
  ip_protocol       = "-1"

  tags = {
    SBO_Billing = "YOUR_TAG_HERE_obi_generative_gui_allow_outgoing"
    Name        = "obi_generative_gui_allow_outgoing"
  }
}
```

obi_generative_gui/security-groups.tf
```diff
@@ -0,0 +1,24 @@
+resource "aws_security_group" "acc_sg" {
+  vpc_id = var.vpc_id
+
+  name        = "main_obi_generative_gui_sg"
+  description = "main security group for obi-generative-gui resources"
+}
+
+resource "aws_vpc_security_group_ingress_rule" "main_subnet_ingress" {
```
danifr: Do we still need this?

obi_one/backend.tf
```diff
@@ -0,0 +1,242 @@
...
+
+  ip_protocol = "tcp"
+  from_port   = var.container_port
+  to_port     = var.container_port
+  cidr_ipv4   = data.aws_vpc.main.cidr_block
+  description = "Allow access to the container port"
+}
+
+
+resource "aws_vpc_security_group_egress_rule" "obi_one_allow_outgoing_tcp" {
```
danifr: OPTINAL, same as before these 2 egress rules can be combined into 1

obi_one/security-groups.tf
```diff
@@ -0,0 +1,24 @@
+resource "aws_security_group" "acc_sg" {
+  vpc_id = var.vpc_id
+
+  name        = "main_obi_one_sg"
+  description = "main security group for obi-one resources"
+}
+
+resource "aws_vpc_security_group_ingress_rule" "main_subnet_ingress" {
```
danifr: Similar to another one, do we still need this? Can we close it down even more?

danifr (changes requested):

obi_one/alb.tf
GianlucaFicarelli: Yeah copy/paste leftover, thanks for spotting it

obi_generative_gui/backend.tf
GianlucaFicarelli: How Name is used? Can we set both SBO_Billing and Name to obi_generative_gui?

danifr: yes, the name is just for humans to be able to be able to identify the rules. You can put anything that you think describes it well

obi_generative_gui/security-groups.tf
GianlucaFicarelli: Both aws_vpc_security_group_egress_rule and aws_vpc_security_group_ingress_rule are defined in backend.tf.
Would it be safe to remove the full security-groups.tf file?

danifr: From a quick grep 'main_subnet_ingress' seems to only be used by the SG `acc_sg`. We can probably delete that SG and all the rules assigned to it. 

I think `obi_generative_gui_ecs_task` and `obi_one_ecs_task` SGs are enough (one per service/task).
Unless there is something else like EFS, DBs, etc. But that can be added later

obi_one/subnets.tf
```diff
@@ -0,0 +1,28 @@
+resource "aws_subnet" "obi_one_ecs_a" {
+  vpc_id            = var.vpc_id
+  availability_zone = "${var.aws_region}a"
+  cidr_block        = "10.0.24.32/27"
```
danifr, line 4: Another thing :) obi_one and gui define the same CIDR bit their subnets a and b

danifr:
All good now, thanks a lot. I am sure there will be new issues that will appear once we merge it but for those we will fix them as they arise.

I will approve it after today's 14h GO/NO-GO meeting.

danifr (changes requested):
Ah another thing that we are missing is `aws_network_acl` definitions for the subnets

danifr (approved):

PR #279, opened 2025-04-25
Upgrade keycloak engine to postgres v16

danifr:
Fixes https://github.com/openbraininstitute/INFRA/issues/152


danifr:
@openbraininstitute/infrastructure I rehearsed the upgrade on a different RDS DB instance and I think it is safer to upgrade to v16 first and then to v17.

This is at least the recommended way from the AWS console. The upgrade process takes around 20 minutes

DriesVerachtert (approved):

PR #282, opened 2025-04-28
Disable Keycloak theme cache everywhere but production

danifr:
For https://github.com/openbraininstitute/INFRA/issues/169


heerener (approved):
Fancy

PR #285, opened 2025-04-29
Add CORS config for entitycore S3 buckets

pgetta:


pgetta:
FYI, @GianlucaFicarelli

pgetta:
It seems there is an issue with CI: `Error acquiring the state lock`. @danifr maybe you can have a look if you have time?

pgetta:
I've tried to restart it couple of times but it keeps failing.

main.tf
```diff
@@ -432,8 +432,10 @@ module "entitycore_svc" {
     ) : (
     "https://staging.openbraininstitute.org/auth/realms/SBO/"
   )
-  s3_bucket_name = var.entitycore_svc_s3_bucket_name
-  image_url      = var.entitycore_svc_image_url
+  s3_bucket_name            = var.entitycore_svc_s3_bucket_name
+  s3_bucket_allowed_origins = ["*"]
```
danifr: shouldn't this be equal to `entitycore_svc_s3_bucket_allowed_origins ` ?

pgetta: Thanks, I've missed that. Fixed now.

danifr (approved):

PR #293, opened 2025-05-07
Increase elasticsearch obp count to 3 nodes

danifr:


DriesVerachtert (commented):
Let's hope that doesn't recreate the cluster :-)

DriesVerachtert (approved):

PR #322, opened 2025-05-21
Use the image in the public AWS ECR repo for Bluenaas, the single cell simulator

DriesVerachtert:


danifr (approved):
I guess we can merge this one. Next time we release in production we can update the production.tfvars

PR #327, opened 2025-05-26
obi-one bumped to 2025.5.3

james-isbister:
* obi-one bumped to 2025.5.3


DriesVerachtert (approved):

WonderPG (approved):
LGTM ! Just one question, why do you guys keep two separate versions (staging and prod) in the `.tfvars` files and don't use the `main.tf` as we do ? By doing so, you are abstracting yourself from the tagging mechanism that CS has in place.

james-isbister:
This is how @GianlucaFicarelli set things up for obi-one @WonderPG 

I'm not sure what the distinction is myself

PR #336, opened 2025-06-05
Split jupyterhub service into 2 different instances

danifr:
One for the main service the other one for Polina's metabolism notebook

### DIFF
--- a/cs/main.tf
+++ b/cs/main.tf
@@ -4,18 +4,45 @@ module "networking" {
   route_table_private_subnets_id = var.route_table_private_subnets_id
 }
 
+# dedicated instance for Polina's Metabolism notebook
+module "jupyterhub_metabolism" {
+  source                            = "./jupyterhub"
+  aws_coreservices_ssh_key_id       = var.aws_coreservices_ssh_key_id
+  vpc_id                            = var.vpc_id
+  private_alb_https_listener_arn    = var.private_alb_https_listener_arn
+  primary_domain                    = var.domain_name
+  jupyterhub_private_subnet         = module.networking.jupyterhub_private_subnet
+  jupyterhub_ec2_type               = "c7i.large"
+  jupyterhub_nginx_port             = 80
+  jupyterhub_port                   = 8080
+  jupyterhub_base_path              = "/jupyterhub_metabolism"
+  jupyterhub_secrets_arn            = var.jupyterhub_secrets_arn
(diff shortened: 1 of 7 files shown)


danifr:
For https://github.com/openbraininstitute/INFRA/issues/206

cs/main.tf
```diff
@@ -4,18 +4,45 @@ module "networking" {
   route_table_private_subnets_id = var.route_table_private_subnets_id
 }
 
+# decicated instance for Polina's Metabolism notebook
```
DriesVerachtert: typo in 'dedicated' :-)

cs/main.tf
```diff
@@ -4,18 +4,45 @@ module "networking" {
...
+  private_alb_https_listener_arn    = var.private_alb_https_listener_arn
+  primary_domain                    = var.domain_name
+  jupyterhub_private_subnet         = module.networking.jupyterhub_private_subnet
+  jupyterhub_ec2_type               = var.jupyterhub_ec2_type
+  jupyterhub_nginx_port             = 80
+  jupyterhub_port                   = 8080
+  jupyterhub_base_path              = "/jupyterhub"
+  jupyterhub_secrets_arn            = var.jupyterhub_secrets_arn
+  jupyterhub_ec2_config_template    = "jupyterhub_config.sh.tpl"
+  jupyterhub_ec2_operating_system   = "ubuntu/images/hvm-ssd-gp3/ubuntu-noble-24.04-amd64-server-20250530"
```
DriesVerachtert, line 41: feel free to ignore this comment, just a suggestion: it could be an option to use a search for the ami instead of a fixed version, like we define some at https://github.com/openbraininstitute/aws-terraform-deployment/blob/staging/amis.tf and then use it like we do at https://github.com/openbraininstitute/aws-terraform-deployment/blob/19983d623903be9a9b6bb280320e8501f74d1ef1/ssh-bastion-vms-on-public-subnets.tf#L3

danifr: My worry is that every time there is a new version and we run terraform the VM will be recreated, or not?

cs/main.tf
danifr: ha! 👀 I will fix it

cs/main.tf
DriesVerachtert: yes that's true: a new VM will be created. We have that issue with the bastion host at the moment.
So yeah if you want to avoid that, then definitely ignore this comment.

DriesVerachtert (approved):

PR #353, opened 2025-06-17
Add small scale simulation service

pgetta:
This deploys a new version of "BlueNaaS", which now supports: single cell, synaptome and small circuit simulations.

Task sizes and autoscaling policies are to be tweaked.


pgetta:
Relates to https://github.com/openbraininstitute/prod-circuit-simulation/issues/42.

small_scale_simulator/dashboard.tf
```diff
@@ -0,0 +1,60 @@
+locals {
+  clustername  = "small-scale-simulator"
+  servicenames = ["redis", "api", "worker"]
+}
+
+resource "aws_cloudwatch_dashboard" "main" {
+  dashboard_name = "BlueNaaS"
```
danifr: is this name correct?

danifr (changes requested):

danifr:
I skimmed through it and it looks correct. It might not work 1st try but I am ok merging and fixing if we find errors

small_scale_simulator/dashboard.tf
pgetta: Ah, good find, thanks, fixed now.

danifr (dismissed):
Cool, approved. I hope the subnets are all free. I have not verified them :D

danifr (approved):
💣

PR #370, opened 2025-06-23
neuroagent version v0.6.2

BoBer78:


ml/agent.tf
```diff
@@ -135,7 +135,7 @@ module "ecs_service_agent" {
...
...titycore-brainatlas-getone|entitycore-brainregion-getall|entitycore-brainregion-getone|entitycore-brainregionhierarchy-getall|entitycore-brainregionhierarchy-getone|entitycore-contribution-getall|entitycore-contribution-getone|entitycore-electricalcellrecording-getall|entitycore-electricalcellrecording-getone|entitycore-emodel-getall|entitycore-emodel-getone|entitycore-etype-getall|entitycore-etype-getone|entitycore-experimentalboutondensity-getall|entitycore-experimentalboutondensity-getone|entitycore-experimentalneurondensity-getall|entitycore-experimentalneurondensity-getone|entitycore-experimentalsynapsesperconnection-getall|entitycore-experimentalsynapsesperconnection-getone|entitycore-memodel-getall|entitycore-memodel-getone|entitycore-mtype-getall|entitycore-mtype-getone|entitycore-organization-getall|entitycore-organization-getone|entitycore-person-getall|entitycore-person-getone|entitycore-reconstructionmorphology-getall|entitycore-reconstructionmorphology-getone|entitycore-species-getall|entitycore-species-getone|entitycore-strain-getall|entitycore-strain-getone|entitycore-subject-getall|entitycore-subject-getone|literature-search-tool|morpho-metrics-tool|web-search-tool"
```
jankrepl: I would definitely use regex instead of enumerating all tools. The below should match all entitycore tools except for the asset download one

```
^entitycore-(?!asset-downloadone$).+
```

BoBer78: I agree this is better, but there are 4 other entitycore tools that are not approved (the two `measurement-annotation` and the 2 `ion-channel`)

jankrepl: ah..sorry..good point..but still I would go for the exclusion pattern.

WonderPG:
```
^(?!.*(downloadone|ionchannelmodel|measurementannotation|plot-generator|resolve-|scs|thumbnail|mcp)).*
```
should do it

WonderPG (approved):

PR #376, opened 2025-06-24
entitycore: bump to 2025.6.5 in prod

GianlucaFicarelli:
This PR will deploy the same version of entitycore in staging to production.
It will require a manual drop of the db in production (there is a short script containing the commands for that)

Still draft, because we are probably going to make a new release before.

PR #381, opened 2025-06-25
Bump entitycore to 2025.6.6

eleftherioszisis:


staging.tfvars
```diff
@@ -58,7 +58,7 @@ hpc_resource_provisioner_scratch_bucket       = ""
 
 entitycore_svc_s3_bucket_name            = "entitycore-data-staging"
 entitycore_svc_s3_bucket_allowed_origins = ["*"]
-entitycore_svc_image_url                 = "public.ecr.aws/openbraininstitute/entitycore:2025.6.5"
+entitycore_svc_image_url                 = "public.ecr.aws/openbraininstitute/entitycore:2025.6.6"
```
GianlucaFicarelli, line 61: We can update also the prod tag in this PR, and close https://github.com/openbraininstitute/aws-terraform-deployment/pull/376 (prod isn't used yet, so we don't need to assess staging before updating the prod tag as it's usually done)
I'll take care of dropping the prod db (that should be empty at the moment)

cc @jdcourcol

danifr:
@eleftherioszisis for the CI to pass, just rebase from `staging`. I just merged https://github.com/openbraininstitute/aws-terraform-deployment/pull/382 and it should be back to normal again :)

staging.tfvars
eleftherioszisis: done

GianlucaFicarelli (approved):

PR #455, opened 2025-07-08
Scale down small scale simulator, optimise provider strategy

pgetta:
This change parameterises capacity provider strategy so that we can exclusively use SPOT instances in staging, while combining SPOT + ON-DEMAND instances in production (so that there is no chance that amazon can recall all workers in prod).


danifr (approved):
thanks!

pgetta:
Thank you, Daniel.

PR #459, opened 2025-07-11
Use multi_az for vlab API prod DB

pgetta:


virtual-lab-manager/db.tf
```diff
@@ -35,11 +35,13 @@ resource "aws_db_instance" "virtual_lab_manager" {
   allocated_storage       = 5    # in gigabytes
   backup_retention_period = 2    # in days
 
+  apply_immediately = false
```
danifr, line 38: Just curious, why this to false?

pgetta: Two things:
* Setting this to false not to have downtime (around 1 -2 minutes) when we merge this, but rather during the maintenance window (some night hours) - potentially less impact for the users.
* Default value for this parameter is false, but I put it explicitly there so that there is no doubt.

danifr: Thanks!

danifr (approved):

pgetta:
Thank you Daniel

PR #469, opened 2025-07-17
Small scale simulator: support multiple worker services

pgetta:
This change makes it possible to have multiple worker services each with a separate size/provider/autoscaling configuration.

Production now uses small Fargate Spot container instances to run single neuron simulations and similar operations (scale wise), while the circuit simulations will run exclusively on Fargate (on-demand).

FYI @eleftherioszisis


pgetta:
The idea is that for the short tasks - various single neuron operations including simulations - we use cheaper **SPOT** instances, while for the longer running stuff - small circuit simulations - to avoid interruptions due to AWS reclaiming some containers - normal **on-demand** ones. This is in **production**.

Staging is using exclusively **SPOT** instances to reduce overall costs.

danifr (approved):

pgetta:
Thank you Daniel

pgetta:
It fails with `Error acquiring the state lock`, is there anything we need to do or just waiting is sufficient to resolve it, @danifr ?

danifr:
I didn't do anything, just rerun the job and this time it worked

pgetta:
Thank you Daniel!

PR #470, opened 2025-07-18
Update health_check in thumbnail_generation_api

GianlucaFicarelli:
This PR can wait after the hotfixes for the other services are deployed to production.

Note also that both /docs and /health are cached by nginx, so the service needs to be updated to bypass the cache when /health is called.


danifr (approved):
LGTM. Up to you if you want to merge it ASAP or wait after for today's deployment to happen. Thanks!

GianlucaFicarelli:
> LGTM. Up to you if you want to merge it ASAP or wait after for today's deployment to happen. Thanks!

It's better to wait after the deployment, and the service should be updated as well with the change for the cache mentioned in the description (still to do).

GianlucaFicarelli:
I'm going to merge this one, since the change in the related PR has been deployed to staging with the latest staging tag.
I checked and the endpoint /health in staging always returns the header x-cached: MISS as expected
https://staging.openbraininstitute.org/api/thumbnail-generation/health

PR #473, opened 2025-07-18
add cloudfront (cdn) for core_webapp

bilalesi:


production.tfvars
```diff
@@ -18,7 +18,7 @@ accounting_svc_docker_image_url           = "public.ecr.aws/openbraininstitute/a
 me_model_analysis_docker_image_url        = "public.ecr.aws/openbraininstitute/me-model-analysis:2025.07.03.1"
 is_nexus_openscience_running              = false
 is_nexus_obp_running                      = true
-jupyterhub_ec2_type                       = "c7i.2xlarge"
+jupyterhub_ec2_type                       = "c7i.16xlarge"
```
danifr: This is not correct. Should be c7i.2xlarge

danifr (changes requested):

production.tfvars
```diff
@@ -27,22 +27,14 @@ bluenaas_task_size = {
   memory = 32768
 }
 
-small_scale_simulator_api_docker_image_url           = "public.ecr.aws/openbraininstitute/single-cell-simulator:api-2025.07.04.1"
```
danifr: hmm another thing, are you sure all these small_scale_simulator related variables need to be deleted?

variables-common.tf
```diff
@@ -134,30 +159,6 @@ variable "small_scale_simulator_worker_task_size" {
   description = "CPU and memory limit for the worker ECS task (number or string format)"
 }
 
-variable "small_scale_simulator_api_task_size" {
```
danifr: Same, this should not be deleted, right?

danifr (changes requested):

variables-common.tf
```diff
@@ -82,12 +82,37 @@ variable "core_web_app_deployment_env" {
...
-  sensitive   = false
 }
 
 variable "core_web_app_next_public_matomo_site_id" {
   type        = string
-  description = "matomo site id <staging|production>"
+  description = "The matomo site id for the core web app"
+}
+
+variable "core_webapp_s3_bucket_name" {
```
danifr, line 87: Where is this bucket being created? needs to be the same repo for all the enviroments?

danifr (changes requested):

production.tfvars
bilalesi: I might missed it in conflicts resolution

variables-common.tf
bilalesi: fixed, conflicts resolution 🙃

production.tfvars
bilalesi: fixed, conflicts resolution 🙃

variables-common.tf
bilalesi: i renamed the vars in `production/staging.tfv` to have the same name `core_webapp_s3_bucket_name` that should give context to terraform ?

danifr: yeah but there is no resource to create that bucket? or does it exist already?

bilalesi: there is `core_webapp/s3.tf` to create it ?

danifr: cool I see it now

staging.tfvars
```diff
@@ -74,3 +74,9 @@ obi_one_task_size = {
   memory = 1024
 }
 obi_generative_gui_docker_image_url = "public.ecr.aws/openbraininstitute/obi-generative-gui:2025.5.4"
+
+# CoreWebApp s3 and CloudFront configuration
+core_webapp_s3_bucket_name                  = "obi-cwa-staging-assets"
+core_webapp_main_s3_bucket_allowed_origins  = ["https://staging.openbraininstitute.org", "https://dev.openbraininstitute.org"]
```
danifr: Do these 3 (core_webapp_main_s3_bucket_allowed_origins, core_webapp_main_cloudfront_aliases, core_webapp_main_cloudfront_certificate_arn) need to be in variables-common?

can just they be module variables an use "var.primary_domain" ?

core_webapp/outputs.tf
```diff
@@ -14,3 +14,38 @@ output "private_lb_rule_suffix" {
   description = "Core Web app Private Loadbalancer Rule Suffix"
   value       = aws_lb_target_group.core_webapp_private.arn_suffix
 }
+
+output "s3_bucket_name" {
```
danifr: Where are you planning to use these output variables from?

danifr (changes requested):

staging.tfvars
bilalesi: good sugg, i removed per env variables, and update `main.tf` to use `primary_domain` both for `s3_bucket_allowed_origins`, `cloudfront_aliases`, one thing need to be done after we have `dev` env to add the domain to `s3_bucket_allowed_origins`

core_webapp/outputs.tf
bilalesi: not needed, removed

danifr: And the other output variables you defined in: https://github.com/openbraininstitute/aws-terraform-deployment/pull/473/files#diff-0f6ba730d79725d9f943fbb7ff769a2845d9cafee07e1ea37c355ec0d536fecc why did you add them there?

bilalesi: as i mentioned here https://github.com/openbraininstitute/prod-virtual-lab/issues/81#issuecomment-3088899628
I only need the `s3_bucke_name`, and the `cloudfront_distribution_id`.
I removed the rest

variables-common.tf
```diff
@@ -91,6 +91,33 @@ variable "core_web_app_next_public_matomo_site_id" {
...
+}
+
+variable "core_webapp_cloudfront_aliases" {
+  type        = list(string)
+  description = "List of domain aliases for core webapp main CloudFront distribution"
+  default     = null
+  sensitive   = false
+}
+
+variable "core_webapp_cloudfront_certificate_arn" {
```
danifr, line 93: this variable is sometimes called `core_webapp_cloudfront_certificate_arn` and in other places `core_webapp_main_cloudfront_certificate_arn`. Please use only one of them

variables-common.tf
```diff
@@ -91,6 +91,33 @@ variable "core_web_app_next_public_matomo_site_id" {
   sensitive   = false
 }
 
+variable "core_webapp_s3_bucket_name" {
+  type        = string
+  description = "S3 bucket name for core webapp main assets"
+  sensitive   = false
+}
+
+variable "core_webapp_s3_bucket_allowed_origins" {
```
danifr: you can remove this variable definiton from the common file

variables-common.tf
```diff
@@ -91,6 +91,33 @@ variable "core_web_app_next_public_matomo_site_id" {
...
+  type        = list(string)
+  description = "Allowed CORS origins for the core webapp main S3 bucket"
+  default     = ["*"]
+  sensitive   = false
+}
+
+variable "core_webapp_cloudfront_aliases" {
```
danifr: same with this one, you can remove it

danifr (changes requested):

variables-common.tf
bilalesi: done

variables-common.tf
bilalesi: done

variables-common.tf
bilalesi: done

core_webapp/variables.tf
```diff
@@ -145,3 +145,31 @@ variable "env_NEXT_PUBLIC_ENABLE_RUN_NOTEBOOK" {
...
+variable "s3_bucket_name" {
+  type        = string
+  description = "Name of the S3 bucket for core webapp assets"
+  sensitive   = false
+}
+
+variable "s3_bucket_allowed_origins" {
+  type        = list(string)
+  description = "Allowed CORS origins for the S3 bucket"
+  default     = ["*"]
```
danifr: I would not like have an open value here. Maybe remove the default value? If a developer wants to test it with '*' it should be a conscious decision rather than the default.

core_webapp/outputs.tf
danifr: hmm I am still thinking about this, what is the advantage of having `s3_bucke_name` as a module output variable? you set the value of the variable in staging.tfvars and production.tfvars.

Maybe I am missing something :D

bilalesi: you're right, removed

danifr (dismissed):
Thanks for addressing all the comments

danifr (approved):

PR #478, opened 2025-07-22
Retire Bluenaas and me-model validation services in staging

pgetta:
These have been superseded by the small scale simulator.


pgetta:
Huh, this turns out a bit more tricky than I thought. Will leave that for now and going to remove modules from both deployment environments once tested in staging.

PR #487, opened 2025-07-25
Cleanup bluenaas and me-model-validation services - step 2

pgetta:


danifr (changes requested):
You can exclude the CODEOWNERS changes, it seems the file has errors and we are actually not using it (it was a half baked implementation inherited from gitlab)

pgetta:
> You can exclude the CODEOWNERS changes, it seems the file has errors and we are actually not using it (it was a half baked implementation inherited from gitlab)

Done, I've reverted the changes to CODEOWNERS file.

danifr (approved):

PR #501, opened 2025-08-04
Remove nexus configuration from virtual_lab

danifr:
to be checked by @bilalesi


bilalesi (approved):
lgtm, 
thank you @danifr

PR #515, opened 2025-08-08
Remove unused variables in ML module

danifr:
I was looking for "nexus" occurrences and I found these 2 variables in your module. I guess we can just delete them cc @BoBer78 @WonderPG @jankrepl


jankrepl (approved):
LGTM:)

PR #516, opened 2025-08-08
Sandbox hpc

heerener:


heerener:
With the S3 resources, we might want to import the existing buckets!

hpc/s3/variables.tf
```diff
@@ -0,0 +1,15 @@
+variable "sboinfrastructureassets_bucket_name" {
```
danifr: Since this is a new brand new variable, shall we avoid using SBO in the name? or you rather keep it like the old name?

sandbox-benchmarks.tfvars
```diff
@@ -0,0 +1,87 @@
...
+nexus_az_letter_id                        = "a"
+core_web_app_docker_image_url             = "bluebrain/sbo-core-web-app:2025.1.0-prod"
+virtual_lab_manager_docker_image_url      = "public.ecr.aws/openbraininstitute/virtual-lab-api:20250226.1"
+thumbnail_generation_api_docker_image_url = "bluebrain/thumbnail-generation-api:latest"
+cell_svc_docker_image_url                 = "public.ecr.aws/openbraininstitute/sonata-cell-position:2025.5.0"
+accounting_svc_docker_image_url           = "public.ecr.aws/openbraininstitute/accounting-service:latest"
+me_model_analysis_docker_image_url        = "public.ecr.aws/openbraininstitute/me-model-analysis:staging"
+is_nexus_openscience_running              = false
+is_nexus_obp_running                      = true
+jupyterhub_ec2_type                       = "c7i.4xlarge"
```
danifr: most of these variables are not used in the sandbox-benchmarks env right?
having an unused c7i.4xlarge Jupyter service sounds expensive :)

hpc/security/compute-cluster.tf
```diff
@@ -168,4 +168,20 @@ resource "aws_security_group" "compute_efs" {
     cidr_blocks = [data.aws_vpc.provided_vpc.cidr_block]
     description = "allow egress within vpc"
   }
+
+  ingress {
```
danifr, line 172: does it need to be so open?

hpc/s3/variables.tf
heerener: Good call; I'll see where else I can rename `sbo{,_}infra` to get rid of the prefix while I'm at it.

sandbox-benchmarks.tfvars
heerener: Jupyter is not set up in this sandbox, but I'll change it just in case of accidents :-)

hpc/security/compute-cluster.tf
heerener: [Yes](https://docs.aws.amazon.com/AWSEC2/latest/UserGuide/efa-start.html#efa-start-security), but this is only for traffic within the security group (that's what the `self` property is for); so only for resources that share the security group.

hpc/main.tf
```diff
@@ -102,9 +102,23 @@ module "resource-provisioner" {
...
 
 module "dynamodb" {
   source        = "./dynamodb/"
   is_production = var.is_production
 }
+
+module "s3" {
+  source                                        = "./s3/"
+  infrastructureassets_bucket_name              = var.infrastructureassets_bucket_name
+  hpc_resource_provisioner_sbo_nexusdata_bucket = var.sbo_nexusdata_bucket
```
danifr: What about this name, can it be changed to something different than nexus? or is it just to much hassle?

danifr (approved):

PR #527, opened 2025-08-12
remove old private alb references

DriesVerachtert:
- **Remove references to old private ALB which isn't used anymore**
- **Remove references to the old private application load balancer**


danifr (approved):
🔥

PR #542, opened 2025-08-19
Remove leading '*' from cloudfront cache_behavior

danifr:


jdcourcol (approved):

bilalesi:
well i think those leading "*" is required because those paths have a version folder, but let"s try it

PR #550, opened 2025-08-22
Configure CORS origins for obi-one (to allow dev env)

pgetta:
This allows dev environment to access staging obi-one deployment as well as tightens up the security a bit as in the default CORS configuration cross-environment access is allowed, see https://github.com/openbraininstitute/obi-one/blob/main/app/config.py#L23 this change enables isolation between production and staging/dev environments.

### DIFF
--- a/main.tf
+++ b/main.tf
@@ -476,7 +476,7 @@ module "obi_one" {
 
   cors_origins = concat(
     ["https://${local.primary_domain}"],
-    var.is_staging ? ["http://localhost:3000", "https://dev.openbraininstitute.org"] : []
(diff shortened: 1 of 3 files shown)


pgetta:
@GianlucaFicarelli just wanted to check with you if you don't have any objections about the change.

pgetta:
Relates to https://github.com/openbraininstitute/INFRA/issues/220.

GianlucaFicarelli (approved):
Looks ok to me, we can do the same also for entitycore

pgetta:
Thank you, @GianlucaFicarelli, I'll prepare the change for the entitycore.

PR #568, opened 2025-09-05
Deploy core web app with hotfixes in prod

pgetta:


danifr (approved):

danifr:
This won't automatically trigger a deployment in prod

pgetta:
Thanks!

I'll wait for Ayima's go to release/build the image and after we can merge this.

pgetta:
Yeah, right, this is just so that we can be a bit faster when we have a go.

PR #581, opened 2025-09-15
Small scale simulator: worker on demand provisioning

pgetta:
This change introduces on demand provisioning for the small scale simulator.

In the current logic the provisioning decision is made based on the job queue length metrics from CloudWatch which is published by the API service.

### DIFF
--- a/main.tf
+++ b/main.tf
@@ -214,7 +214,9 @@ module "small_scale_simulator" {
   keycloak_server_url = "https://${local.primary_domain}/auth/"
 
   api_task_size = var.small_scale_simulator_api_task_size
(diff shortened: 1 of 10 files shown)


pgetta:
Related to https://github.com/openbraininstitute/Ultraliser/issues/54.

danifr (approved):
LGTM. do you need me to add a new valid redirect URL to the small scale simulator client?

pgetta:
This is pure backend, all the workers communicate with the client via API available thru existing domains, so no changes to Keycloak needed, thanks!

I'm pretty sure it's not going to work right away, but let's see.

PR #582, opened 2025-09-16
Remove explicit env var declaration for AWS_REGION in lambda fn

pgetta:
It is already provided by the AWS runtime environment.


danifr (approved):
Thanks Pavlo!

pgetta:
Relates to https://github.com/openbraininstitute/Ultraliser/issues/54.

PR #592, opened 2025-09-17
initial version of the launch-service

mgeplf:


mgeplf:
No need to review yet.

launch_server/variables.tf
```diff
@@ -0,0 +1,120 @@
...
+  description = "Admin username for the database"
+  type        = string
+}
+
+variable "az_subscription_id" {
+  description = "Subscription ID in Azure"
+  type        = string
+}
+
+variable "azure_client_id" {
```
mgeplf: @heerener do you think I should make `azure_client_id`, `az_subscription_id`, `azure_tenant_id` secrets too, so that they don't show up in github?

launch_server/variables.tf
```diff
@@ -0,0 +1,120 @@
...
+  description = "region in Azure"
+  type        = string
+}
+
+variable "az_batch_pool_name" {
+  description = "batch pool name in Azure"
+  type        = string
+}
+
+variable "az_upload_blob_sas_url" {
```
mgeplf: this should definitely be a secret, I will change it

heerener (dismissed):

heerener (approved):

PR #594, opened 2025-09-18
Add small_scale_simulator workers variables

danifr:
Not used in prod yet but current code makes them mandatory


pgetta:
@danifr thank, I've pushed a small change, what has been called workers has become "daemon" workers, I've adjusted the config accordingly.

DriesVerachtert (approved):

PR #595, opened 2025-09-19
Point entitycore to the restored database

heerener:


heerener:
No longer needed, restore test was a success.

PR #602, opened 2025-09-24
Small scale sim improvements

pgetta:
Changes:
* Add billing tags
* Use Python 3.12 in provisioner lambda
* Use smaller statistics window for job queue length metrics


danifr (approved):

pgetta:
THanks!

PR #606, opened 2025-09-26
Small scale simulator: fix aggregation period

pgetta:
@danifr  it turns out I didn't commit new name for the lambda function.
Also the 15 sec aggregation period is not supported:

`Error: An error occurred (InvalidParameterValue) when calling the GetMetricStatistics operation: The parameter Period must be a value in the set [1, 5, 10, 20, 30].`

So these are the additional changes.


danifr (approved):
never easy man 😅

PR #630, opened 2025-10-08
obi_one_v2: Add EFS storage for perf test

GianlucaFicarelli:


obi_one_v2/efs.tf
```diff
@@ -0,0 +1,32 @@
...
+
+resource "aws_efs_mount_target" "test_perf_efs" {
+  file_system_id  = aws_efs_file_system.test_perf_efs.id
+  subnet_id       = aws_subnet.obi_one_v2.id
+  security_groups = [aws_security_group.efs.id]
+}
+
+resource "aws_security_group" "efs" {
+  name        = "efs-sg"
+  description = "Allow NFS traffic"
```
mgeplf, line 16: NFS?

mgeplf (approved):

obi_one_v2/efs.tf
GianlucaFicarelli: EFS implements the NFS protocol under the hood AFAIK, so the port is the same used for NFS. I can update the comment to be more specific for EFS, though

mgeplf: ok, good to know that's what I thought was true; I just found it strange. No need to change.

PR #644, opened 2025-10-13
Replace obi-one with obi-one-v2, remove EFS

GianlucaFicarelli:
- rename /obi-one to /obi-one-v1 (to be removed completely in a separate PR)
- rename /obi-one-v2 to /obi-one
- add missing application env variables
- use t3-small in staging (can be increased if needed)
- add local cache to mount-s3 (limited by the size of EC2: ~ 19 GB usable for the cache)
- remove EFS volume and mount


james-isbister:
Thanks for this. Ok to go ahead on my side, but I'll leave it to someone else to assess the code!

DriesVerachtert (approved):

PR #645, opened 2025-10-13
Entitycore bucket versioning

heerener:


DriesVerachtert (approved):

eleftherioszisis:
I assume there is some prior discussion as to why this needs to be enabled now, right?

heerener:
> I assume there is some prior discussion as to why this needs to be enabled now, right?

There is, see https://github.com/openbraininstitute/INFRA/issues/280

eleftherioszisis (approved):
Thanks!

PR #647, opened 2025-10-14
Add read-only access for Athena to RDS database of the virtual lab manager

DriesVerachtert:


athena_ro_access_to_rds_db/main.tf
```diff
@@ -0,0 +1,200 @@
...
+  name_prefix = "${var.name_prefix}_ro_access_to_db"
+
+  assume_role_policy = <<-EOT
+  {
+    "Version": "2012-10-17",
+    "Statement": [
+      {
+        "Action": "sts:AssumeRole",
+        "Principal": {
+          "Service": "lambda.amazonaws.com"
```
heerener, line 16: should it be lambda or athena?

DriesVerachtert: What this PR does, is creating some lambda with a jdbc driver which acts as some kind of 'athena adapter': it allows athena to do queries in some datasource through an adapter, which is in this case an rds database.

heerener: Makes sense; there's a lot that seems to go through lambda, just wanted to make sure it wasn't a copy/paste :-)

heerener (approved):

PR #653, opened 2025-10-14
obi-one version bump

james-isbister:


GianlucaFicarelli (approved):
Not a big issue because the deployment to production is not automatic, but the production tag should be updated only after staging has been successfully tested.

PR #654, opened 2025-10-14
Automatically refresh EC2 when the template is updated

GianlucaFicarelli:
Based on the note in https://registry.terraform.io/providers/hashicorp/aws/latest/docs/resources/autoscaling_group#instance_refresh

> A refresh will not start when version = "$Latest" is configured in the launch_template block. To trigger the instance refresh when a launch template is changed, configure version to use the latest_version attribute of the aws_launch_template resource.


danifr (approved):
good catch

PR #671, opened 2025-10-22
feat: add auth manager service

bilalesi:
- auth manager store tokens in the db


DriesVerachtert (dismissed):

auth-manager/db.tf
```diff
@@ -0,0 +1,50 @@
+resource "aws_db_subnet_group" "auth_manager_db_cluster_subnet_group" {
+  name       = "auth-manager-db-cluster-group"
+  subnet_ids = [aws_subnet.auth_manager_db_a.id, aws_subnet.auth_manager_db_b.id]
+}
+
+data "aws_secretsmanager_secret_version" "auth_manager_database_password" {
+  secret_id = var.auth_manager_secrets_arn
```
danifr, line 9: We need a new secret entry for this right? We just need to remember to add it

bilalesi: yes, that's right

bilalesi:
@danifr 
the auth manager, need probably new client or re-using existing one, i don't have a preference, you know better
also both the client ID, and UUID as envs i put placeholders for the moment

bilalesi:
there is also env that should be added to the secret you can see how here 
https://github.com/openbraininstitute/auth-manager/blob/main/auth-manager/README.md
envs:
AUTH_MANAGER_TOKEN_VAULT_ENCRYPTION_KEY
ACK_STATE_SECRET
+
DB password

danifr:
Depends on https://github.com/openbraininstitute/aws-terraform-deployment-common/pull/105

staging.tfvars
```diff
@@ -121,3 +121,6 @@ obi_one_v2_ecs_task_size = {
 
 # CoreWebApp s3 and CloudFront configuration
 core_webapp_s3_bucket_name = "core-webapp-static-assets-staging"
+auth_manager_svc_image_url = "985539765147.dkr.ecr.us-east-1.amazonaws.com/auth-manager:2025.10.28.1"
+keycloak_client_uuid       = "589eff2d-4d3a-4003-a91f-bed9ec463ef1"
```
danifr: Why do you need to client uuid as a variable?

bilalesi: We have this endpoint to keycloak, to get offline sessions of specific user,
probably will remove it (need to discuss with jdc), I will update terraform in that case
```
url = "{}/admin/realms/{}/users/{}/offline-sessions/{}".format(
            self.config.issuer, self.config.realm, user_id, self.config.client_uuid
        )
```

danifr: alright, if this is needed maybe in the future I can give access to your API user to query client details too. So with the client name you can get the uuid. But we can set it as a variable for now. Thanks

bilalesi: thanks, probably will not need this endpoint, I will confirm soon
btw: the client need to have service account Role (with `manage-realm`, `manage-users` roles)
for the moment, i don't think we can revoke user sessions, but JDC said it's possible, I will also discuss with him about it, but it's also will be clearer when we start testing the service in staging

another thing related to KC:
in the client we should have this URLs added as allowed redirect urls 
`https://staging.openbraininstitute.org/api/auth-manager/v1/offline-token/callback` and 
`https://staging.openbraininstitute.org/app/consent-feedback`

and both for `preview, dev` prefixes, 
same should be applied when deploy to prod

danifr: great thanks for the info, adding them now

danifr: All done, for prod I only added:
- https://www.openbraininstitute.org/api/auth-manager/v1/offline-token/callback and 
- https://www.openbraininstitute.org/app/consent-feedback

bilalesi:
@danifr, if everything looks good on your end (and the PR doesn’t raise any red flags), shall we move it to staging?

auth-manager/alb.tf
```diff
@@ -0,0 +1,43 @@
...
+    path     = "${var.root_path}/health"
+    protocol = "HTTP"
+  }
+
+  tags = var.auth_manager_svc_tags
+}
+
+resource "aws_lb_listener_rule" "auth_manager_private_listener_rule" {
+  listener_arn = var.private_alb_listener_arn
+  priority     = 610
```
danifr: This one seems to be taken by entitycore, please use 612 for example

auth-manager/db.tf
```diff
@@ -0,0 +1,52 @@
...
+  engine_version              = "17"
+  allow_major_version_upgrade = true
+  multi_az                    = true
+  instance_class              = "db.t3.small"
+
+  deletion_protection = true #tfsec:ignore:AVD-AWS-0177
+  allocated_storage   = 10   # in gigabytes
+
+  backup_retention_period = 14 # in days
+  backup_window           = "04:00-05:00"
```
danifr: can you change it to `"01:00-02:00"` this is what we have for the other DBs. It was recently changes by @heerener

auth-manager/service.tf
```diff
@@ -0,0 +1,334 @@
...
+  security_group_id = aws_security_group.auth_manager_ecs_task.id
+
+  ip_protocol = "tcp"
+  from_port   = 8000
+  to_port     = 8000
+  cidr_ipv4   = data.aws_vpc.main.cidr_block
+  description = "Allow port 8000 http"
+}
+
+resource "aws_vpc_security_group_ingress_rule" "auth_manager_allow_in_tcp" {
```
danifr, line 53: Do we need this to be that open?

auth-manager/service.tf
```diff
@@ -0,0 +1,334 @@
...
+
+      portMappings = [
+        {
+          hostPort      = 8000
+          containerPort = 8000
+          protocol      = "tcp"
+        }
+      ]
+
+      healthcheck = {
```
danifr: If there is no proper check in place I'd remove this block

auth-manager/service.tf
```diff
@@ -0,0 +1,334 @@
...
+    Statement = [
+      {
+        Effect = "Allow"
+        Action = [
+          "logs:CreateLogGroup",
+          "logs:CreateLogStream",
+          "logs:DescribeLogStreams",
+          "logs:PutLogEvents",
+        ]
+        Resource = "arn:aws:logs:*:*:*"
```
danifr: Better to close this one a bit: `arn:aws:logs:${var.aws_region}:${data.aws_caller_identity.current.account_id}:log-group:${aws_cloudwatch_log_group. auth_manager_ecs_task_logs.name}*`

you need to add a data block to get the aws_caller_identity

auth-manager/subnets.tf
```diff
@@ -0,0 +1,36 @@
...
+  tags = {
+    Name = "auth_manager_db_b"
+  }
+}
+
+
+resource "aws_subnet" "auth_manager_ecs_a" {
+  vpc_id            = var.vpc_id
+  availability_zone = "${var.aws_region}a"
+  cidr_block        = "10.0.23.32/27"
```
danifr: used by entitycore

auth-manager/subnets.tf
```diff
@@ -0,0 +1,36 @@
...
+  tags = {
+    Name = "auth_manager_ecs_a"
+  }
+}
+
+resource "aws_subnet" "auth_manager_ecs_b" {
+  vpc_id            = var.vpc_id
+  availability_zone = "${var.aws_region}b"
+  cidr_block        = "10.0.23.64/27"
```
danifr: same, already in use

danifr (changes requested):

auth-manager/service.tf
bilalesi: yep i think it's fine, it's only within the vpc

auth-manager/service.tf
bilalesi: done

auth-manager/alb.tf
bilalesi: done

auth-manager/db.tf
bilalesi: done

auth-manager/subnets.tf
bilalesi: done

auth-manager/subnets.tf
bilalesi: done

danifr (approved):
Cool! thanks for addressing my comments. We can merge and fix on the go if there is something not working

bilalesi:
thanks @danifr, merging then

PR #680, opened 2025-10-24
obi_one: set protect_from_scale_in = false

GianlucaFicarelli:
Set `protect_from_scale_in = false` to allow faster scale down of the EC2 instances after a new deployment.

It has been observed that after a new deployment of the service, that requires spinning up a new EC2 instance, the desired capacity of the autoscaling group stays at 2, although it decreases to 1 after some time, and the old EC2 instance is then destroyed.

This is an attempt to make the scaling down faster.

cc @danifr


DriesVerachtert (approved):

danifr (approved):
Let's try it!

PR #701, opened 2025-10-30
Temporarily point entitycore to the restored database.

heerener:


heerener:
No longer needed, restore test was successful.

PR #702, opened 2025-10-30
fix keycloak issuer url for auth-manager svc

bilalesi:


bilalesi:
I will wait for @heerener green light

danifr (approved):
ah I missed this one...

PR #733, opened 2025-11-10
Add launch-system to staging

GianlucaFicarelli:
Based on the branch https://github.com/openbraininstitute/aws-terraform-deployment/tree/sandbox-nse

Notes:
- For consistency with the name of the git repository, the names launch/launch-service/launch-system have been renamed to launch-system.
- There are 3 ECS clusters (api, orchestrator, executor), each one with a task definition (api, orchestrator, default_executor). In this way they should be more isolated with regards to logging and config in general, but we could decide to aggregate them differently.
- There are 2 subnets, trusted and untrusted. The untrusted network should have rules to forbid the connection to internal resources (db, redis, other services), while it should allow to call at least the internal launch-system api, and the entitycore api, and access the EFS volume that will be defined separately. At the moment no restriction is defined, and need to verify if and how these constraints can be satisfied with the current infrastructure.
- I tried to simplify the names of the resources by removing the type in the name, since they should be more readable and it's considered a good practice ("Don't repeat the resource type in the resource name") for example in
  - https://docs.aws.amazon.com/prescriptive-guidance/latest/terraform-aws-provider-best-practices/structure.html#naming-conventions
  - https://docs.cloud.google.com/docs/terraform/best-practices/general-style-structure
  - https://www.terraform-best-practices.com/naming


launch_system/db.tf
```diff
@@ -17,35 +20,40 @@ resource "aws_db_instance" "launch" {
...
   allow_major_version_upgrade = true
   multi_az                    = true
-  instance_class              = "db.t4g.micro"
+  instance_class              = var.db_instance_class
 
   deletion_protection = true #tfsec:ignore:AVD-AWS-0177
-  allocated_storage   = 50   # in gigabytes
+  allocated_storage   = var.db_allocated_storage
 
   backup_retention_period = 14 # in days
```
heerener: Since you're adding it to the backup plan, there's no need for db-level backups (see https://github.com/openbraininstitute/INFRA/issues/301 - this should have been disabled for the launch db already).

launch_system/api.tf
```diff
@@ -148,9 +148,29 @@ resource "aws_ecs_task_definition" "api" {
...
-          name  = "LAUNCH_SERVER_URL" # for backward compatibility
+          name  = "LAUNCH_SERVER_URL" # deprecated, use LAUNCH_SYSTEM_API_URL
           value = var.launch_system_api_url
         },
+        {
+          name  = "ACCOUNTING_URL"
+          value = var.accounting_url
+        },
+        {
+          name  = "AUTH_MANAGER_URL"
```
mgeplf, line 159: thanks, I should have notified you about this.

GianlucaFicarelli: np, this list can be checked with the possible variables in config.py but we'll need to try it

launch_system/elasticache.tf
```diff
@@ -0,0 +1,43 @@
...
+    cidr_blocks = [var.vpc_cidr_block]
+    description = "allow access from within VPC"
+  }
+
+  tags = merge(var.tags, { Name = "launch_system_redis" })
+}
+
+# Create ElastiCache subnet group with private subnets
+resource "aws_elasticache_subnet_group" "redis" {
+  name = "launch_system_redis_subnet_group"
```
heerener:
```suggestion
  name = "launch-system-redis-subnet-group"
```

When deploying manually I saw this error; this change should fix it: 

```
│ Error: creating ElastiCache Subnet Group (launch_system_redis_subnet_group): operation error ElastiCache: CreateCacheSubnetGroup, https response error StatusCode: 400, RequestID: 039acfd9-143c-473a-82ee-27307eb959ae, api error InvalidParameterValue: The parameter CacheSubnetGroupName is not a valid identifier. Ide
ntifiers must begin with a letter; must contain only ASCII letters, digits, and hyphens; and must not end with a hyphen or contain two consecutive hyphens.
```

GianlucaFicarelli: Fixed in the last commit, thanks for spotting the error.
I find quite annoying that some names and ids don't allow the underscore as a valid character, while others allow it. It seems inconsistent, do you know if there is a clear rule?

heerener: Unfortunately "it's inconsistent" is as far as I got as well.

main.tf
```diff
@@ -804,34 +805,64 @@ module "virtual_lab_manager" {
...
+module "launch_system" {
+  source = "./launch_system"
+
+  count = var.is_staging ? 1 : 0
 
   aws_region                    = local.aws_region
   vpc_id                        = local.vpc_id
+  account_id                    = local.account_id
   private_alb_listener_arn      = local.private_alb_https_listener_arn
   internet_access_route_id      = local.route_table_private_subnets_id
```
heerener, line 824: This is somewhat confusing: the ID for the internet access route is the ID of the route table for the private subnets? Shouldn't that be `local.route_table_public_id`?

GianlucaFicarelli: This comes from the initial setup and I didn't notice it. However, I see that also the other services are using the same `local.route_table_private_subnets_id` and according to the [output](https://github.com/openbraininstitute/aws-terraform-deployment-common/blob/1f8625f2385d4d94d99865281c96126d57f9019d/outputs.tf#L43), `route_table_private_subnets_id` is the "ID of the route table for all private subnets in the main region", while `route_table_public_id` is the "ID of the route table for all public subnets in the main region". 
This is then used in `aws_route_table_association` to associate the route_table_id with the (private) subnet_id. Why should we use the route table for the public subnets?

heerener: I'm mostly judging based on the name of the variable and didn't really dig into what it's being used for. If it's not for public access, then renaming it might make it clearer. I guess this is a legacy artifact that was once intended for public use and then changed :-)

GianlucaFicarelli: Clarified offline, `internet_access_route_id` has probably been named in this way because the private networks can connect to the outside internet thanks to this route table.

eleftherioszisis (dismissed):

heerener (approved):

PR #749, opened 2025-11-13
auth-manager v5 2025.11.14.1

bilalesi:


danifr (dismissed):
Approved. I guess this one can be merged at any time since it is not really used at the moment. In any case, please check with Ayima before merging. Thanks!

danifr (approved):

PR #750, opened 2025-11-14
Shared efs

heerener:
Implementation for https://github.com/openbraininstitute/prod-platform-architecture/issues/151


sandbox-hpc.tfvars
```diff
@@ -22,6 +22,8 @@ notebook_service_bucket_name              = "obi-notebook-service-statistics-san
 notebook_service_k8s_thread_enabled       = false
 notebook_service_accounting_enabled       = false
 
+neuroagent_image_tag = "neuroagent-v0.11.4"
```
danifr, line 26: Maybe for the future but it would be great to find a better solution for the sandbox envs. Something like: inherit the variables values from staging.tfvars and overwrite them from the ones set up in sandbox-hpc.tfvars. 

So in the `sandbox-<env>tfvars` files we only to set up the variables we need to change.

public_data_efs/datasync.tf
```diff
@@ -0,0 +1,91 @@
...
+          "arn:aws:s3:::${var.entitycore_open_data_bucket}/*"
+        ]
+      }
+    ]
+  })
+}
+
+resource "aws_datasync_location_s3" "internal_source" {
+  s3_bucket_arn = "arn:aws:s3:::${var.entitycore_open_data_bucket}"
+  subdirectory  = "/public"
```
GianlucaFicarelli, line 50: From the [docs](https://registry.terraform.io/providers/hashicorp/aws/latest/docs/resources/datasync_location_s3#subdirectory-1) subdirectory is the prefix, but in s3 the the prefix is `public` without a leading `/`. Does it work also with `/`?

public_data_efs/datasync.tf
```diff
@@ -0,0 +1,91 @@
...
+    mtime                  = "PRESERVE"
+    uid                    = "INT_VALUE"
+    gid                    = "INT_VALUE"
+    posix_permissions      = "PRESERVE"
+    preserve_devices       = "NONE"
+    bytes_per_second       = -1 # unlimited
+  }
+
+  schedule {
+    schedule_expression = "cron(0 0 ? * SUN *)"
```
GianlucaFicarelli: We may want to tune the schedule later, but for start I would set once per day.

public_data_efs/datasync.tf
heerener: Good catch - I tried with the code like this and saw my synchronized data show up ind the container. If it turns out that something is going wrong this will be the first culprit!

public_data_efs/datasync.tf
heerener: Done!

sandbox-hpc.tfvars
heerener: I would very much like that, because people keep forgetting to update the `sandbox-*tfvar` files 😅

public_data_efs/datasync.tf
```diff
@@ -0,0 +1,92 @@
...
+    uid                    = "INT_VALUE"
+    gid                    = "INT_VALUE"
+    posix_permissions      = "PRESERVE"
+    preserve_devices       = "NONE"
+    bytes_per_second       = -1 # unlimited
+  }
+
+  schedule {
+    # minute | hour | day of month | month | day of week | year
+    schedule_expression = "cron(0 0 * * * *)"
```
GianlucaFicarelli: According to this [doc](https://docs.aws.amazon.com/eventbridge/latest/userguide/eb-scheduled-rule-pattern.html) 

> You can't use * in both the Day-of-month and Day-of-week fields. If you use it in one, you must use ? in the other.

We could also add a comment to this doc to be aware of these peculiarities.

launch_system/executor.tf
```diff
@@ -68,6 +69,17 @@ resource "aws_ecs_task_definition" "default_executor" {
...
+      mountPoints = [
+        {
+          containerPath = "/data/aws_s3_internal/public",
+          sourceVolume  = "public-internal-data"
+        },
+        {
+          containerPath = "/data/aws_s3_open",
+          sourceVolume  = "public-open-data"
+        }
+      ]
```
eleftherioszisis, line 81: It could be declared somewhere else, but just to check, these should be mounted readonly. Is it the case here?

heerener: I noticed there were a couple of files missing from the PR while trying to answer this question 😬 

Check out the `aws_efs_access_point`s defined in `public_data_efs/main.tf` and let me know whether it's good enough!

public_data_efs/datasync.tf
heerener: Done!

launch_system/executor.tf
eleftherioszisis: I see. Yeah, I think it looks good

GianlucaFicarelli (dismissed):

eleftherioszisis (dismissed):

GianlucaFicarelli (approved):

PR #753, opened 2025-11-18
increase projects per virtual lab

bilalesi:


bilalesi:
@danifr this PR is urgent, should be merged to stg, test it and go to prod (request from @jdcourcol )

danifr (approved):

PR #755, opened 2025-11-18
allow auth-manager deployment only in stg

bilalesi:


danifr (changes requested):
@bilalesi Hmm now that I am looking at this:
```
- module.auth_manager.aws_subnet.auth_manager_db_a
- module.auth_manager.aws_subnet.auth_manager_db_b
- module.auth_manager.aws_subnet.auth_manager_ecs_a
- module.auth_manager.aws_subnet.auth_manager_ecs_b
```

Deleting and recreating subnets it is always messy and usually requires manual intervention.

Instead of using count at the modele level maybe we can do it at the task.
Lemme know I can take care of the changes. Thanks

danifr:
@bilalesi I replaced your commit with another one. I think this way it is a bit cleaner and does not require manual intervention or recreating the resources. 

When we want to have it running in prod we just need to set that `number_of_containers` to 1, or simply revert this commit

bilalesi:
@danifr thanks, i cannot approve the PR though (i created it)

DriesVerachtert (approved):

danifr (approved):

PR #764, opened 2025-11-21
Use neuroagent image from OBI ECR repo

danifr:


jankrepl (approved):
LGTM:) Thank you @danifr

PR #795, opened 2025-11-28
Core web app: clean up redeployment role for staging

pgetta:
As it's no longer used because we've switched to versioned tags.


danifr (approved):
Thanks!

PR #811, opened 2025-12-04
Core web app: put CDN in front of image optimization endpoint

pgetta:
This is essentially to serve all optimised images via CDN.

The change is described in https://github.com/openbraininstitute/prod-virtual-lab/issues/95#issuecomment-3595918107.


danifr (approved):
No idea if this will work but terraform-wise the code makes sense so I will approve it. Thanks!

pgetta:
Same here, just praying to TF gods.

pgetta:
Thanks!

pgetta:
Impressive, seems to be working as expected.

PR #813, opened 2025-12-04
Increase optimized image cache TTL for Core web app CDN

pgetta:


pgetta:
CDN caching is working so great that I've decided to increase TTL (the content doesn't change often) to keep it that way.

danifr (approved):
Nice!

pgetta:
Thanks!

PR #822, opened 2025-12-09
Block all public access for static_storage S3 bucket

danifr:
Fix for one of the critical findings SecurityHub raised: "S3 general purpose buckets should block public read access"

https://github.com/openbraininstitute/INFRA/issues/361

### DIFF
--- a/static-server/main.tf
+++ b/static-server/main.tf
@@ -96,13 +96,13 @@ resource "aws_s3_bucket_metric" "static_storage_metrics" {
   name   = "EntireBucket"
 }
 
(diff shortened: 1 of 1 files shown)


pgetta:
I wonder, does this have an impact on the static server?

I've checked the bucket and we have some probably important files in there (this is staging):
Screenshot 2025-12-09 at 11 57 04

pgetta:
Hm, actually we will need to revise those files. I'm pretty sure the sitemap is outdated.

heerener (approved):

PR #831, opened 2025-12-12
Upgrade auth manager to 2025.12.12.1 in staging

pgetta:
Auth manager is not used neither in production nor in dev/staging. So this change has no impact on the current production deployment.


DriesVerachtert (approved):

pgetta:
Thanks!

PR #838, opened 2025-12-17
fix obi-one-v2 dashboard

jdcourcol:
- current dashboard for obi-one is empty , fixing service name


obi_one_v2/dashboard.tf
```diff
@@ -1,6 +1,6 @@
 locals {
   clustername = "obi_one_v2_ecs_cluster"
-  servicename = "obi_one_v2_ecs_service"
+  servicename = "obi_one_v2_service"
```
GianlucaFicarelli, line 3: Alternatively we could rename the aws_ecs_service in container.tf, but I don't have a strong preference.

GianlucaFicarelli (approved):

obi_one_v2/dashboard.tf
jdcourcol: me neither, I just want it to work.

jdcourcol:
@DriesVerachtert @heerener , the error is :
" **Error: No valid credential sources found**
│ 
│ Please see https://developer.hashicorp.com/terraform/language/backend/s3
│ for more information about providing credentials.
│ 
│ Error: failed to refresh cached credentials, no EC2 IMDS role found,
│ operation error ec2imds: GetMetadata, failed to get API token, operation
│ error ec2imds: getToken, http response error StatusCode: 400, request to
│ EC2 IMDS failed"

I don't see what this has to do with the change.

danifr:
it fails because the change comes from a forked repo, and that repo does not have the right credentials set up to be able to run the CI.

Since the change is small, I can simply merge it and bypass the CI requirements.

danifr:
Change applied correctly in:  https://github.com/openbraininstitute/aws-terraform-deployment/actions/runs/20913095465/job/60080046600

PR #931, opened 2026-01-26
Enable readonlyRootFilesystem for thumbnail-generation service

danifr:
I checked the code and it seems it needs write access to `/tmp`, `/var/cache/nginx`, and `/app/output`

### DIFF
--- a/thumbnail-generation-api/thumbnail-generation-api-container.tf
+++ b/thumbnail-generation-api/thumbnail-generation-api-container.tf
@@ -153,6 +153,10 @@ resource "aws_ecs_task_definition" "thumbnail_generation_api_task_definition" {
   }
 
   volume {
+    name = "var-run"
+  }
+
+  volume {
     name = "app-output"
   }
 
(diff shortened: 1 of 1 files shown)


thumbnail-generation-api/thumbnail-generation-api-container.tf
```diff
@@ -144,14 +144,40 @@ resource "aws_ecs_task_definition" "thumbnail_generation_api_task_definition" {
...
+          },
+          {
+            sourceVolume  = "var-cache-nginx",
+            containerPath = "/var/cache/nginx"
+          },
+          {
+            sourceVolume  = "app-output",
+            containerPath = "/app/output"
+          }
+        ],
```
bilalesi, line 188: we are using matplotlib for generating plots, and it  needs a writable config/cache, so probably we need to allow that too (`/.config/matplotlib`), or add this env
```py
environment = [
  { name = "MPLCONFIGDIR", value = "/tmp/matplotlib" }
]
```

we used also supervisor and it generate `/var/run` but i don't know if this is what `readonlyRootFilesystem` will require

danifr: Thanks!!! Comments addressed in https://github.com/openbraininstitute/aws-terraform-deployment/pull/931/commits/7ddb1588346c477790a1f8fd82454dedbb922a5b

bilalesi: thanks @danifr 
hopefully will work as expected

bilalesi (approved):

PR #939, opened 2026-01-26
Enable readonlyRootFilesystem for virtual-lab manager service

danifr:

AmazonQ mentioned that we only need `/tmp` mount for Python/uvicorn temporary operations.


bilalesi (approved):
yes that's right, we don't use file system, other than `uv`, i don't think we don't have others

PR #951, opened 2026-01-28
Obi1 version increment

dkeller9:


danifr (commented):
I let @james-isbister approve this one.

james-isbister:
All good! Now we only update the staging.tfvars @dkeller9 

And then @danifr copies the values from this file to the production.tfvars as part of the push to production

So we should revert your change to the production.tfvars please

dkeller9:
I changed production.tfvars back to the original.

james-isbister (approved):

danifr:
Thanks @dkeller9 !

PR #975, opened 2026-02-02
Enable readonlyRootFilesystem for neuroagent service

danifr:
Depends on https://github.com/openbraininstitute/neuroagent/pull/683

Before merging this one a new version of neuroagent needs to be released and deployed in staging.


WonderPG:
As mentioned in https://github.com/openbraininstitute/neuroagent/pull/683, I explicitly needed to provide execute right in docker-compose to this `/tmp` folder. I am not sure if this is done by default or is something more needs to be added in the config. We need to be able to execute scripts located within this folder.

WonderPG (approved):

PR #989, opened 2026-02-04
Fix core web app preview deployments

pgetta:
This is an attempt to address an issue with Amplify not having access to Route53 to control subdomains. As well as some minor improvements.


danifr (approved):
let's try!

PR #1011, opened 2026-02-06
VLab API: increase log retention to 14 days

pgetta:
Log retention was set to 5 days which is too short for any not-immediately-visible issue to be analysed.

Relates to [Increase log retention period](https://github.com/openbraininstitute/core-web-app/issues/1345).


DriesVerachtert (approved):

pgetta:
Thank you Dries

PR #1018, opened 2026-02-09
Enable readonlyRootFilesystem for obi-one service

danifr:

@GianlucaFicarelli if I am not mistaken obi-one writes to `/tmp` but that dir is already mounted as en external tmpfs volume.


GianlucaFicarelli:
> if I am not mistaken obi-one writes to `/tmp` but that dir is already mounted as en external tmpfs volume.

That's correct, /tmp is mounted as tmpfs and it can be used for writing. However:
- it's quite limited (512 MB) compared to the space available in the container (30 GB iirc)
- I'm not sure that all the code writes in /tmp only, @james-isbister and @chr-pok could you confirm if that's the case?

If 512 MB aren't enough, can we make a bigger /tmp without tmpfs?

chr-pok:
> * I'm not sure that all the code writes in /tmp only, @james-isbister and @chr-pok could you confirm if that's the case?

Not sure exactly. For circuit/connectivity metrics for example, the circuit is downloaded into `tempfile.TemporaryDirectory()` (in case mounting is not available), not sure if this is related to `/tmp`.

james-isbister:
`I'm not sure that all the code writes in /tmp only, @james-isbister and @chr-pok could you confirm if that's the case?`

It should be the case, yes. I checked the endpoints I'm less familiar with and that does seem to be the case.

GianlucaFicarelli:
> Not sure exactly. For circuit/connectivity metrics for example, the circuit is downloaded into `tempfile.TemporaryDirectory()` (in case mounting is not available), not sure if this is related to `/tmp`.

When using `tempfile.TemporaryDirectory()` without specifying a directory:

> If dir is not None, the file will be created in that directory; otherwise, a default directory is used. The default directory is chosen from a platform-dependent list, but the user of the application can control the directory location by setting the TMPDIR, TEMP or TMP environment variables

Note that there is also the env variable `OUTPUT_DIR` that is used in some parts of the code. It's set to `/tmp` in https://github.com/openbraininstitute/obi-one/blob/207401d68def0dbf4b0c7eda0507fa86b183ba0f/Dockerfile#L91 and not overridden in aws-terraform-deployment.

With all the considerations above, I think we can merge this PR, and test the endpoints that require write access.

If the storage of 512 MB isn't enough, we can replace tmpfs with a local storage (probably slower, but bigger).

---

Side note, I noticed that the /tmp directory in staging contains a few small morphologies (4 swc morphologies, 3.1 MB each, created yesterday between 15:20 and 16:23). It means that they are not correctly cleaned up, so it should be verified what part of the code is responsible for that so it can be fixed. 
I opened https://github.com/openbraininstitute/obi-one/issues/577 for this.

GianlucaFicarelli (approved):

PR #1025, opened 2026-02-10
Allow core web app preview role stop Amplify jobs

pgetta:
So that CI can stop currently running builds for previous commits.


danifr (approved):
gogogog

PR #1036, opened 2026-02-11
Enable readonlyRootFilesystem for launch-system API service

danifr:
another one for https://github.com/openbraininstitute/INFRA/issues/391


GianlucaFicarelli (approved):
LGTM for the api, but I think we can do the same for the orchestrator.
Not sure it can be done for the executor, since it needs to install the venv at runtime.

GianlucaFicarelli (approved):

PR #1042, opened 2026-02-11
Set env varibales to make sure service work with readonlyRootFilesystem

danifr:
Explanation from AmazonQ:

> The code uses pynwb which creates cache directories. The issue is that pynwb (and its dependency hdmf) try to create cache directories in /tmp/.cache/pynwb-hdmf/ but the directory doesn't exist and the library can't create it due to permission issues.


danifr:
Pavlo set them it at the container level. No need to do it via terraform.

PR #1101, opened 2026-02-26
Fix the s3 bucket CORS for preview envs

BoBer78:
We had an issue with the preview branch. The plots where getting CORSed because of our s3 bucket not allowing the new : 
`[branch-name].preview.openbraininstitute.org/)` pattern


danifr (approved):
thanks!

PR #1102, opened 2026-02-26
Disable datasync schedules to Azure

heerener:


danifr (approved):
I think it makes sense

PR #1105, opened 2026-02-26
Add an archive for SES events in eventbridge to figure out when/what events are created

DriesVerachtert:
https://github.com/openbraininstitute/INFRA/issues/355


danifr (approved):
🦤

PR #1147, opened 2026-03-06
entitycore: increase db migration timeouts

GianlucaFicarelli:
The deployment to staging fails for a timeout, so increasing:

- DB_MIGRATION_STATEMENT_TIMEOUT_MS: 30000 -> 60000
- DB_MIGRATION_LOCK_TIMEOUT_MS: 5000 -> 10000

Error:
> sqlalchemy.exc.OperationalError: (psycopg2.errors.QueryCanceled) canceling statement due to statement timeout

from https://us-east-1.console.aws.amazon.com/ecs/v2/clusters/entitycore_ecs_cluster/services/entitycore_ecs_service/tasks/b990b45b344f4162ba8bd7e20a504c55/logs?region=us-east-1

### DIFF
--- a/entitycore_svc/backend.tf
+++ b/entitycore_svc/backend.tf
@@ -164,6 +164,14 @@ resource "aws_ecs_task_definition" "entitycore_ecs_definition" {
           name  = "API_ASSET_POST_MAX_SIZE"
           value = var.api_asset_post_max_size
         },
+        {
+          name  = "DB_MIGRATION_STATEMENT_TIMEOUT_MS"
+          value = var.db_migration_statement_timeout_ms
+        },
+        {
+          name  = "DB_MIGRATION_LOCK_TIMEOUT_MS"
+          value = var.db_migration_lock_timeout_ms
+        },
       ]
 
       secrets = [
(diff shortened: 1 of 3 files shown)


DriesVerachtert (approved):

GianlucaFicarelli:
For the record, the slow query in new deployment was successful after 53 seconds, see [here](https://us-east-1.console.aws.amazon.com/ecs/v2/clusters/entitycore_ecs_cluster/services/entitycore_ecs_service/tasks/fc3e7443b29c48b89a9639d1f97aeaf7/logs?region=us-east-1)

However, in the meantime the health check marked the instance as not healthy, so it was stopped AFTER the migration.
This left the db updated, but the old application running for a few minutes, until a new deployment was tried, and succeeded quickly since the db migration was done already.

So at the end the deployment is successful, but we might want to improve it to avoid a possible disruption.

Notes:
- The query that will be executed during the deployment to production should be faster, since the asset table is smaller
- In this case, the changes to the db were backward compatible, so they didn't cause issues to the old application

cc @eleftherioszisis @mgeplf

PR #1159, opened 2026-03-10
launch-system: move `launch_system_network` back into `launch_system`

mgeplf:


heerener (changes requested):
Deleting / recreating networks in terraform is usually a pain because it never seems to clean up all resources that have network interfaces in the affected networks, at least not before trying the delete.

Better to try with `moved` blocks; I hope that will also avoid the need to recreate the datasync tasks and lose our history on those.

mgeplf:
> Better to try with moved blocks; 

I have the moved blocks in moved.tf; did they not work?

heerener (approved):
> > Better to try with moved blocks;
> 
> I have the moved blocks in moved.tf; did they not work?

I don't know what the ~~rebase~~ merge has to do with it, but it seems like now terraform does take them into account. It should indeed be enough to have them somewhere in the root of the repo. Latest output looks a lot better!

I'll let @DriesVerachtert have the final say, but for me it looks okay.

DriesVerachtert (approved):

PR #1198, opened 2026-03-17
Wire AWS PCS into the launch_system

mgeplf:
* `small` and `large` nodes
* `SLURM_SECRET` was added to the SECRETS already

### DIFF
--- a/launch_system/compute_cell_definitions.tf
+++ b/launch_system/compute_cell_definitions.tf
@@ -1,7 +1,8 @@
-# All compute cell definitions in one place. Cell A uses this module's resources;
-# other cells are static configuration.
+# All compute cell definitions in one place
 
 locals {
+  slurmrestd_endpoint = [for e in awscc_pcs_cluster.cluster.endpoints : e if e.type == "SLURMRESTD"][0]
+
   compute_cell_definitions = {
     cell_a = {
       vendor     = "aws"
@@ -35,7 +36,20 @@ locals {
             task_family     = aws_ecs_task_definition.inait_executor.family
           }
         ]
-        cluster = []
+        cluster = {
+          type                 = "cluster"
+          username             = "obiuser"
+          uid                  = 4000
+          gid                  = 4000
+          homedir              = "/data/scratch/obiuser"
+          slurm_url            = "http://${local.slurmrestd_endpoint.private_ip_address}:6820/slurm/v0.0.43"
+          slurm_accounting_url = "http://${local.slurmrestd_endpoint.private_ip_address}:6820/slurmdb/v0.0.43"
+          slurm_secret         = "$${SECRET:SLURM_SECRET}"
+          instance_types = {
+            small = awscc_pcs_queue.pcs_queue_small.name
+            large = awscc_pcs_queue.pcs_queue_large.name
+          }
+        }
       }
     }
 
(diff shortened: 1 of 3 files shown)


heerener (dismissed):

launch_system/pcs.tf
```diff
@@ -65,20 +65,60 @@ resource "aws_launch_template" "pcs_launch_template" {
   }
 }
 
-resource "awscc_pcs_compute_node_group" "pcs_nodegroup" {
+resource "awscc_pcs_compute_node_group" "pcs_nodegroup_small" {
   name       = "cluster-nodegroup"
```
GianlucaFicarelli: Shouldn't we use different names for resources pcs_nodegroup_small and pcs_nodegroup_large?

launch_system/pcs.tf
GianlucaFicarelli, line 204: Not changed in this PR, but fsx_s3 role and policy seem unused

launch_system/compute_cell_definitions.tf
```diff
@@ -35,7 +36,20 @@ locals {
...
         ]
-        cluster = []
+        cluster = {
+          type                 = "cluster"
+          username             = "obiuser"
+          uid                  = 4000
+          gid                  = 4000
+          homedir              = "/data/scratch/obiuser"
+          slurm_url            = "http://${local.slurmrestd_endpoint.private_ip_address}:6820/slurm/v0.0.43"
+          slurm_accounting_url = "http://${local.slurmrestd_endpoint.private_ip_address}:6820/slurmdb/v0.0.43"
```
GianlucaFicarelli, line 46: Minor, maybe we can move `v0.0.43` to a variable so that the version is more explicit?
From my understanding it should be supported by the scheduler version `25.05` defined in awscc_pcs_cluster.
From https://slurm.schedmd.com/release_notes.html I see there are already new versions 25.11 and 0.0.44, we could consider using the latest ones since it's the first deployment?

launch_system/pcs.tf
mgeplf: hmm, so it's likely working because of something else; any ideas @heerener?

launch_system/pcs.tf
mgeplf: yes, fixed

launch_system/compute_cell_definitions.tf
mgeplf: it's PCS that provides the version: https://docs.aws.amazon.com/pcs/latest/userguide/slurm-versions.html and we're on the latest one.
I have added a note to that respect in the `resource "awscc_pcs_cluster" "cluster"`

launch_system/pcs.tf
heerener: Nice catch, I didn't realise!

There is an IAM role called `AWSServiceRoleForFSxS3Access_fs-0800d3c2d84785425` that is used for the sync between FSx and S3. I see no reference to it in the terraform code, so my best guess is that it got auto-generated when creating the DRA.

Based on all of this I'd suggest trying to remove the role and policy here to see if anything breaks.

mgeplf: > Nice catch, I didn't realise!

I'll change this in a different PR.

GianlucaFicarelli (approved):

PR #1209, opened 2026-03-19
Try unsetting slurm purge while accounting is disabled

heerener:


heerener:
Unsetting it will just force it back to -1, so this is not a solution.

PR #1236, opened 2026-03-26
Unify launch_system and launch_system_network, remove count

GianlucaFicarelli:
- Unify launch_system and launch_system_network
- remove count -> launch_system[0] is moved to launch_system

Similar to https://github.com/openbraininstitute/aws-terraform-deployment/pull/1159, but it also removes count from launch_system so it's always deployed to production as well.

Depends on https://github.com/openbraininstitute/aws-terraform-deployment/pull/1233


launch_system/subnets.tf
```diff
@@ -1,3 +1,51 @@
+resource "aws_subnet" "trusted_a" {
+  vpc_id            = var.vpc_id
+  availability_zone = "${var.aws_region}a"
+  cidr_block        = "10.0.29.0/28"
```
danifr, line 4: I think this is gonna fail, we might need to free and delete the existing subnets 1st and then create this ones on the same ranges. We can check together tomorrow.

GianlucaFicarelli: Good point, I was expecting that the resource is moved because of moved.tf, but we can check terraform plan.
If not working, an alternative could be to allocate new ranges for cidr_block.
If it's simpler, I can also merge this PR with https://github.com/openbraininstitute/aws-terraform-deployment/pull/1233

danifr: Ah well yes, we can try, maybe it won't be that bad

danifr (approved):

PR #1245, opened 2026-03-30
launch_system: turn on slurm accounting

mgeplf:
* To be able to see the status of jobs that have run, one needs to have accounting enabled
* by default, on PCS, slurm has `MinJobAge=300`; so after 5 minutes the information is gone


mgeplf:
FYI: This might take a while to apply the first time, but it should be stable once it's merged

danifr (approved):

PR #1266, opened 2026-04-07
Clean up unused Core Web app dev instance

pgetta:
Dev instance is no longer used after we had switched to preview deployments. This PR is to clean that up.


danifr (approved):
nice thanks for the clean up!

pgetta:
Thanks Daniel!

PR #1276, opened 2026-04-09
Core Web app preview deployments: use staging keycloak client

pgetta:
That is needed to make preview deployment work with the auth-manager (Keycloak client should match) so that we can test submitting jobs which require offline tokens.


pgetta:
This is related to https://github.com/openbraininstitute/Ultraliser/issues/197.

danifr (approved):

PR #1328, opened 2026-04-23
Neuroagent TypeScript - first deployment

jankrepl:


DriesVerachtert:
@jankrepl just curious, why are you using different cidr ranges for ml_typescript_subnet_a_cidr and the b_cidr for staging, prod and hpc sandbox?

jankrepl:
> @jankrepl just curious, why are you using different cidr ranges for ml_typescript_subnet_a_cidr and the b_cidr for staging, prod and hpc sandbox?

Good point. That is not necessary. Will remove that. This is WIP

What is necessary though is for a fixed environment (e.g. staging) to make sure that the cidr ranges of neuroagent/neuroagent-typescript and all the other services not to collide.

main.tf
```diff
@@ -373,6 +374,47 @@ module "ml" {
...
 }
 
+module "ml_typescript" {
+  source = "./ml"
+
+  aws_region   = local.aws_region
+  account_id   = local.account_id
+  instance_key = var.ml_typescript_instance_key
+
+  is_production   = var.is_production
```
WonderPG, line 384: Is this used ?

WonderPG (commented):
I understand the approach to duplicate neuroagent from the one module we created, but isn't this approach meant for actually keeping both deployments in the future ? Wouldn't it be easier to do `cp -r ml/ ml-ts/` and prepend `ts-` to every `agent` string within the folder, such that we can just delete `ml/` as soon as we switch to the ts version for good ? I understand this is not standard best practice, just wondering if it's worth the effort.

main.tf
jankrepl: It seems to be used for some backups (not a mechanism we introduced)


```

⏺ Yes — it's used in ml/agent.tf:25 to conditionally add a backup tag to the ECS service when in production:

  tags = var.is_production ? { obi_backup_plan = var.obi_backup_plan } : {}

  And declared in ml/variables.tf:17.
```

WonderPG (dismissed):

WonderPG (dismissed):

WonderPG (dismissed):

staging.tfvars
```diff
@@ -25,7 +25,11 @@ keycloak_url_with_auth = "https://staging.cell-a.openbraininstitute.org/auth/"
 
 neuroagent_docker_image_url = "985539765147.dkr.ecr.us-east-1.amazonaws.com/neuroagent:neuroagent-v0.17.2"
 
-grading_service_docker_image_url = "985539765147.dkr.ecr.us-east-1.amazonaws.com/grading-service:latest"
+neuroagent_typescript_docker_image_url = "985539765147.dkr.ecr.us-east-1.amazonaws.com/neuroagent:neuroagent-typescript-v0.1.1"
+ml_neuroagent_typescript_bucket_name   = "ml-neuroagent-typescript-staging"
+ml_typescript_subnet_a_cidr            = "10.0.5.0/24"
```
danifr: Do we need these variables in staging and production? Since they have the same values and they are not meant to be changed, they just add unnecessary lines.

jankrepl: Yes, we need them. We're instantiating the same module a second time, so we must provide any values that differ from defaults or from the first instance. The subnet CIDRs must be unique per module instance to avoid conflicts, and the bucket name differs between staging and production. 

Note that we will have 2 instances only temporarily and if this new typescript version works well we will remove the old one and then I can remove these variables.

jankrepl: Ah..but now I am realizing that there is a CIDR conflict already with the grading service. See my latest commit that fixed it. d3ada18

WonderPG (approved):

staging.tfvars
danifr: alright, maybe this can be cleanup in the future then, thanks for the answer

PR #1336, opened 2026-04-28
Enable auth manager in production

pgetta:


danifr (approved):
I am not a big fan of default values in variables but I think for this case it works :)

PR #1356, opened 2026-05-05
Manage Core Web app preview branches and domains from GH CI

pgetta:
For the preview deployments it turns out that the model where :
* AWS manages Amplify branches - auto creation/deletion in sync with git branches
* Github CI manages deployments

In reality doesn't work well for us, as we have substantial amount of non-PR git branches, which due to auto-sync are counted towards the quota.

Multiple times we had reached the quota (50 branches) with only ~24 PRs.

I'm making updates to set a ground for disabling auto-sync and let Github CI manage also Amplify branches to address the issue.


danifr (approved):
alright sounds good

pgetta:
Thank you Daniel

PR #1367, opened 2026-05-06
Core Web app preview: fix GH deploy role ACLs

pgetta:
It seems for `Get/UpdateDomainAssociation` the IAM authorization is evaluated against the app resource (apps/APP_ID) - the app ARN without a trailing segment - not against the domain sub-resource.


danifr (approved):
nice!

PR #1368, opened 2026-05-06
Fix auth-amanger hardcoded variable values

danifr:
Remove unused variables, fix hardcoded env and redirect domain


pgetta (approved):

bilalesi (approved):
thanks @danifr

PR #1407, opened 2026-05-19
Initial setup for internal entitycore data sync

heerener:
The S3 bucket already exists, so we can create the DataSync objects complete with new policy that allows cross-account access

On the S3 bucket, new policy is prepared but will not be applied yet; this can only happen once the datasync role has been set up. It will require a separate commit to apply the bucket policy to the bucket by setting the `source_datasync_role` variable.


DriesVerachtert (approved):
🐶

PR #1424, opened 2026-05-26
launch-system: PCS hpc7a.96xlarge nodes are only in `use1-az6`

mgeplf:


mgeplf:
The subnet needs to be `use1-az6` to be able to access `hpc7a.96xlarge`.

By luck, we're already using this in staging, however, I'm not sure what happens when this gets applied to prod - I assume it will work, and only impact the PCS subnet?

@jdcourcol can confirm that this is necessary to support the IT'IS folks, so it's important.

However, it's probably useful to have people who know AWS chime in: ie @heerener / @danifr / @DriesVerachtert ?

danifr:
yeah I think it makes sense

danifr (approved):

heerener (approved):
Discussed on Teams, if it works by specifying the AZ like this that would be cool and avoid any issues with the random mapping of AZs to letter suffixes: https://docs.aws.amazon.com/ram/latest/userguide/working-with-az-ids.html

jdcourcol:
yes, I confirm.

mgeplf:
Now this needs to be cherry picked to production; can @heerener or @danifr do that?

PR #1449, opened 2026-06-04
Add support for UDP port forwarding from the NLB to Juanjo's VM

DriesVerachtert:
https://github.com/openbraininstitute/INFRA/issues/558

### DIFF
--- a/development_vm/udp_port_forward.tf
+++ b/development_vm/udp_port_forward.tf
@@ -1,0 +1,55 @@
+# Only enabled if public_nlb_arn is set, to forward UDP traffic.
+
+resource "aws_lb_target_group" "udp_forward_to_vm" {
+  count = var.public_nlb_arn == null ? 0 : 1
+
+  name        = replace("${var.vm_name}-udp", "_", "-")
+  port        = var.udp_port_forward_from_public_nlb
+  protocol    = "UDP"
+  target_type = "instance"
+  vpc_id      = var.vpc_id
+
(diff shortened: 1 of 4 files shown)


development_vm/udp_port_forward.tf
```diff
@@ -0,0 +1,55 @@
...
+    interval            = 30
+  }
+}
+
+resource "aws_lb_target_group_attachment" "vm" {
+  count = var.public_nlb_arn == null ? 0 : 1
+
+  target_group_arn = aws_lb_target_group.udp_forward_to_vm[0].arn
+  target_id        = aws_instance.instance.id
+  port             = 8443
```
danifr: you can use var.udp_port_forward_from_public_nlb here

development_vm/udp_port_forward.tf
```diff
@@ -0,0 +1,55 @@
...
+}
+
+
+resource "aws_lb_listener" "udp_listener" {
+  count = var.public_nlb_arn == null ? 0 : 1
+
+  load_balancer_arn = var.public_nlb_arn
+  port              = 8443
```
danifr: and here too

danifr (changes requested):

development_vm/udp_port_forward.tf
DriesVerachtert: should be ok now!

danifr (approved):

PR #1454, opened 2026-06-05
Allow incoming UDP 8443 traffic in the network ACL of the public subnets that are used by the public network load balancer

DriesVerachtert:
https://github.com/openbraininstitute/INFRA/issues/558

### DIFF
--- a/subnets-public.tf
+++ b/subnets-public.tf
@@ -37,6 +37,15 @@ resource "aws_network_acl" "public" {
     from_port  = 443
     to_port    = 443
   }
+  # Allow port 8443 from anywhere for Juanjo's VM
+  ingress {
+    protocol   = "udp"
+    rule_no    = 111
+    action     = "allow"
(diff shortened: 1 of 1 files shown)


subnets-public.tf
```diff
@@ -37,6 +37,15 @@ resource "aws_network_acl" "public" {
     from_port  = 443
     to_port    = 443
   }
+  # Allow port 8443 from anywhere for Juanjo's VM
+  ingress {
+    protocol   = "udp"
+    rule_no    = 111
+    action     = "allow"
+    cidr_block = "0.0.0.0/0"
+    from_port  = 8443
```
heerener, line 46: shouldn't this be the ephemeral range?

heerener (approved):

subnets-public.tf
DriesVerachtert: It's udp, not tcp :-P

PR #1499, opened 2026-06-22
Grading service: inject web launch URL env var

pgetta:


pgetta:
Related to [avoid copy pasting token for moodle exercises.](https://github.com/openbraininstitute/PMO/issues/652).

danifr (approved):

PR #1513, opened 2026-06-23
Disable transit_encryption in EFS

GianlucaFicarelli:
Disable transit_encryption in EFS to test performances when reading data in launch-system executors.

If there is an improvement, it may require some additional changes to limit access.
If there aren't improvements, it can be reverted.

Draft because it cannot be merged and deployed to staging before staging is unfrozen.

See https://github.com/openbraininstitute/prod-build-circuit/issues/28#issuecomment-4778063994 for more context


heerener (approved):

GianlucaFicarelli:
Closing as not needed/planned anymore

PR #1535, opened 2026-06-30
Add self-service redeploy role + outputs for grading-service

pgetta:
## What

Enables the **grading-service** CI (`openbraininstitute/grading-service`) to roll out its own ECS deployments without a per-release Terraform apply. This is the one-time, infra-side half of the self-service release contract.

- **Redeploy IAM role** — instantiate `github_ecs_redeploy_role` for grading-service, **unconditionally** (no `count`/`is_staging` gate) so the role is created in **both** the staging and production accounts. Real tenants release to prod; the obi test tenant to staging. The role trusts only `repo:openbraininstitute/grading-service:*` via OIDC and can do nothing but force a new deployment of this one service (`ecs:UpdateService` / `DescribeServices` / `DescribeTaskDefinition` / `DescribeClusters`, scoped to the single service/cluster/task-def). No image push, no other service, no cross-repo access.
  - Role name: `gh_redeploy_ecs_api_grading-service`.
- **Outputs** — export `ecs_cluster_name` / `ecs_service_name` / `ecs_task_definition_name` from the `grading_service` module, and `grading_service_redeploy_role_arn` from the root, so the team can wire the ARN into their GitHub Environments.

## Why unconditional (differs from notebook-service / keycloak roles)

The existing redeploy roles are `count = var.is_staging ? 1 : 0` because they only redeploy in staging. grading-service releases to **both** environments, so the role (and its output) must exist in both accounts — hence no gate, and the output references the module directly (no `[0]` index).

## Blast radius

None to running infrastructure. `terraform plan` shows only the new IAM role + inline policy and the new outputs — **no change** to the grading-service ECS service or task definition.

## Expected role ARNs (one per account)

- staging: `arn:aws:iam::992382665735:role/gh_redeploy_ecs_api_grading-service`
- production: `arn:aws:iam::671250183987:role/gh_redeploy_ecs_api_grading-service`

## Sequencing

This is **step 1** of 2. It is safe to merge/apply now and does not touch the running service. The image flip to the rolling `:staging` / `:prod` alias is a separate, **held** PR that must only merge **after** the grading-service CI has pushed those aliases to ECR.

### DIFF
--- a/grading_service/outputs.tf
+++ b/grading_service/outputs.tf
@@ -2,3 +2,15 @@ output "private_lb_rule_suffix" {
   description = "Grading service load balancer rule suffix"
   value       = aws_lb_target_group.main.arn_suffix
 }
+
+output "ecs_cluster_name" {
+  value = aws_ecs_cluster.main.name
+}
+
+output "ecs_service_name" {
+  value = aws_ecs_service.api.name
+}
+
+output "ecs_task_definition_name" {
+  value = aws_ecs_task_definition.api.family
+}
--- a/main.tf
+++ b/main.tf
@@ -388,6 +388,21 @@ module "github_keycloak_ecs_redeploy_role" {
   ecs_task_definition_name = module.cs.keycloak_ecs_task_definition_name
 }
 
+module "github_grading_service_ecs_redeploy_role" {
+  source = "./github_ecs_redeploy_role"
+
+  # Unconditional (no count): grading-service deploys to BOTH staging and
+  # production, so this role is needed in both accounts (unlike the staging-only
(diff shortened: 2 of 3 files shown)


DriesVerachtert (approved):

pgetta:
Thank you Dries.

I'll request an approval from JDC to merge this despite the staging freeze (because it's an isolated change) or just wait if I don't get it.

Edit: we want to have that rather sooner than later.

pgetta:
Relates to [Tenant friendly deployment pipeline](https://github.com/openbraininstitute/PMO/issues/688).

PR #1536, opened 2026-06-30
Point grading-service image at rolling alias (:staging / :prod)

pgetta:
Merge this **only after** the grading-service CI has pushed the rolling `:staging` and `:prod` aliases to ECR. Applying before the alias exists would stall the ECS rollout — Fargate cannot pull a non-existent tag (old tasks keep serving, but the deployment hangs). Kept as a draft until then.

## What

Switch the grading-service ECS task def from the pinned date tag to the rolling alias that CI moves on each release:

- `staging.tfvars` → `…/grading-service:staging`
- `production.tfvars` → `…/grading-service:prod`

## Why no drift afterwards

Once flipped, the image string is a constant alias. `terraform plan` shows **no diff** even after CI re-points the underlying digest, because the task-def `image` attribute never changes. CI's `aws ecs update-service --force-new-deployment` restarts tasks against the same task-def revision (Fargate always pulls fresh) and never creates a Terraform-visible change. From here, releases roll out via CI with no per-release Terraform apply.

## Depends on

- #1535 (redeploy role + outputs) — merge & apply first.
- One maintainer release run from grading-service CI to create the `:staging` / `:prod` aliases in ECR.

## Apply order

1. Merge #1535, apply in both accounts, hand the two role ARNs to the grading-service team.
2. Team configures GitHub and does one maintainer release run → aliases land in ECR.
3. Mark this PR ready, merge, apply in both accounts.


pgetta:
Relates to [Tenant friendly deployment pipeline](https://github.com/openbraininstitute/PMO/issues/688).

DriesVerachtert (approved):

PR #1538, opened 2026-06-30
Fix PCS resource creation in sandbox-nse

GianlucaFicarelli:
Fixes errors when deploying `module.launch_system` PCS resources in the sandbox-nse environment:

### Changes

- **Fix invalid S3 bucket name**: `local.suffix` produced an empty string for non-staging/non-production environments, resulting in a trailing hyphen (`obi-pcs-scratch-fsx-`). Now uses `var.deployment_env` as suffix.
- **Make PCS large instance type configurable**: Extracted the hardcoded `hpc7a.96xlarge` into a new `pcs_large_instance_type` variable, allowing sandbox to use a smaller instance type (`c5n.large`).
- **Make EFA optional for large node group**: Added `pcs_large_enable_efa` variable to switch between EFA and non-EFA launch templates. EFA requires specific (large/expensive) instance types not needed in sandbox.
- **Fix FSx Data Repository Association timeout**: Increased create/delete timeout from 10m to 20m to accommodate slower provisioning.
- **Add `replace_triggered_by` on PCS queues**: Ensures queues are automatically replaced when their associated node groups change (e.g. AMI or instance type update), preventing `ResourceConflict` errors during apply.

### Environment settings

| Variable | sandbox-nse | staging/production |
|----------|------------|-------------------|
| `pcs_large_instance_type` | `c5n.large` | `hpc7a.96xlarge` |
| `pcs_large_enable_efa` | `false` | `true` |

### DIFF
--- a/sandbox-nse.tfvars
+++ b/sandbox-nse.tfvars
@@ -135,4 +135,6 @@ azure_nfs_internal_public_data_path                = "/obibatchnfsstg/publicdata
 resource_provisioner_container_hash                = "1837363b2413d5e7cfd8c7a859ba4b3cd35c685acd4240f70101364aaaeee0b8"
 resource_provisioner_container_uri                 = "985539765147.dkr.ecr.us-east-1.amazonaws.com/hpc-resource-provisioner:0.5.13.dev8"
 
-pcs_ami = "ami-0d196e97f4b5b69b4"
+pcs_ami                 = "ami-0fa11a6354572d8b5"
+pcs_large_instance_type = "c5n.large" # 2 vCPUs, 5.3 GiB
+pcs_large_enable_efa    = false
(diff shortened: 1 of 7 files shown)


sandbox-nse.tfvars
```diff
@@ -135,4 +135,6 @@ azure_nfs_internal_public_data_path                = "/obibatchnfsstg/publicdata
 resource_provisioner_container_hash                = "1837363b2413d5e7cfd8c7a859ba4b3cd35c685acd4240f70101364aaaeee0b8"
 resource_provisioner_container_uri                 = "985539765147.dkr.ecr.us-east-1.amazonaws.com/hpc-resource-provisioner:0.5.13.dev8"
 
-pcs_ami = "ami-0d196e97f4b5b69b4"
+pcs_ami                 = "ami-0fa11a6354572d8b5"
```
GianlucaFicarelli, line 138: This is the AMI ID of neurodamus-1779890699, available in the sandbox but different from staging and production.
If needed, the same ID should be made available.

mgeplf: afaik, we don't make the AMI public, so I think they can't be used across `account_id` without doing permissions or tokens or something.  ATM we copy them, so that's why the IDs are different

mgeplf (approved):

eleftherioszisis (approved):

PR #1539, opened 2026-06-30
Bump obi-one to v2026.6.29

james-isbister:
Bumps obi-one image tag in staging.tfvars from 2026.6.28 to 2026.6.29 (released from openbraininstitute/obi-one @ main).


danifr (approved):

danifr (changes requested):
wait no, staging is frozen, we need to wait unless this is a bug fix release only

james-isbister:
Good stop ;) But this is an agreed bug fix in ultraliser we highlighted as necessary with @evakennyobi and @marwan-abdellah this morning so we are good to merge it

danifr (approved):
alrighty!

PR #1556, opened 2026-07-03
Wire LTI 1.3 config into grading-service ECS task

pgetta:
## Context

The `grading-service` app migrated from **LTI 1.1** (OAuth1 / shared-secret) to **LTI 1.3 Advantage + AGS** (commit `2c507cb` in the grading-service repo, using `pylti1p3`). LTI 1.3 changes the runtime contract in two ways the current ECS task definition does **not** yet satisfy — the tasks would crashloop on the new image without them.

## What this changes

- **`PUBLIC_BASE_URL` (env)** — the tool's externally reachable HTTPS base URL, used to build the OIDC login/launch/jwks URLs handed to the platform. Set to `https://${local.cell_a_primary_domain}` → `https://[staging.]cell-a.openbraininstitute.org`; combined with the existing `base_path = /api/grading-service`. The app **requires** it and fails startup otherwise.
- **`LTI_TOOL_PRIVATE_KEY` (secret)** — RSA private key (PEM) that signs AGS client-credentials assertions and backs `/lti/jwks`. Read from a new key in the **existing** `grading_service` secret; the current IAM policy already grants `GetSecretValue` on the whole secret ARN, so no IAM change.

Per-tenant Moodle registrations (`client_id`/`deployment_ids`/issuer/URLs) are baked into the image via the tenant submodules, so there are **no new per-platform env vars/secrets**, and LTI 1.3's asymmetric keys mean there's no Moodle client secret to inject.

1. **deployment-common:** add a stable `LTI_TOOL_PRIVATE_KEY` (RSA PEM) key to the `grading_service` Secrets Manager secret. Its RFC 7638 thumbprint is the JWKS `kid`, so it must not change across redeploys.


pgetta:
Relates to https://github.com/openbraininstitute/PMO/issues/687.

DriesVerachtert (approved):

PR #1570, opened 2026-07-07
launch-system: access private project data in PCS

GianlucaFicarelli:
It uses mount-s3 (not S3 Files) with a cache.

### DIFF
--- a/launch_system/cloud-init.cfg
+++ b/launch_system/cloud-init.cfg
@@ -32,5 +32,55 @@ runcmd:
   - mkdir -p /data/aws_s3_internal
   - mount -t efs -o tls,accesspoint=${opendata_access_point_id} ${publicdata_efs_id}:/ /data/aws_s3_open
   - mount -t efs -o tls,accesspoint=${publicdata_access_point_id} ${publicdata_efs_id}:/ /data/aws_s3_internal
+  - mkdir -p /data/aws_s3_internal/private
+
+write_files:
+  - path: /etc/fuse.conf
+    permissions: '0644'
+    content: |
+      user_allow_other
+
+  - path: /usr/local/bin/prolog-mount-private-data.sh
+    permissions: '0755'
+    content: |
+      #!/bin/bash
+      set -euo pipefail
+      # Slurm Prolog: mount private data for the job.
+      # Comment format: s3|bucket_name|prefix|mount_path
+      COMMENT=$(scontrol show job "$SLURM_JOB_ID" | grep -oP 'Comment=\K[^\s]+' || true)
+      if [ -z "$COMMENT" ] || [[ "$COMMENT" != s3|* ]]; then
+        exit 0
+      fi
+      IFS='|' read -r _ BUCKET PREFIX MOUNT_PATH <<< "$COMMENT"
+      if [ -z "$BUCKET" ] || [ -z "$PREFIX" ] || [ -z "$MOUNT_PATH" ]; then
+        exit 0
+      fi
+      mkdir -p "$MOUNT_PATH"
+      mount-s3 "$BUCKET" "$MOUNT_PATH" \
+        --prefix "$PREFIX/" \
+        --read-only \
+        --region ${region} \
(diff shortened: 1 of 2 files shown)


launch_system/cloud-init.cfg
```diff
@@ -32,5 +32,55 @@ runcmd:
...
+    permissions: '0644'
+    content: |
+      user_allow_other
+
+  - path: /usr/local/bin/prolog-mount-private-data.sh
+    permissions: '0755'
+    content: |
+      #!/bin/bash
+      set -euo pipefail
+      # Slurm Prolog: mount private data for the job.
```
mgeplf: I'd add something like `mount private data for the job based off Slurm comment`

launch_system/cloud-init.cfg
```diff
@@ -32,5 +32,55 @@ runcmd:
...
+      # Comment format: s3|bucket_name|prefix|mount_path
+      COMMENT=$(scontrol show job "$SLURM_JOB_ID" | grep -oP 'Comment=\K[^\s]+' || true)
```
mgeplf: can't one use `SLURM_JOB_COMMENT`: https://slurm.schedmd.com/prolog_epilog.html#:~:text=SLURM_JOB_COMMENT%20Comment%20added%20to%20the%20job.%20Available%20in%20Prolog%2C

instead of scontrol and piping?

launch_system/cloud-init.cfg
```diff
@@ -32,5 +32,55 @@ runcmd:
...
+      if [ -z "$COMMENT" ] || [[ "$COMMENT" != s3|* ]]; then
+        exit 0
```
mgeplf, line 95: emit warning that comment is missing? I wonder where it get's reported

launch_system/cloud-init.cfg
GianlucaFicarelli: Done

launch_system/cloud-init.cfg
GianlucaFicarelli: Done

launch_system/cloud-init.cfg
GianlucaFicarelli: Done by redirecting to a separate log file, since stdout and stderr in prolog/epilog aren't logged.

GianlucaFicarelli: From https://slurm.schedmd.com/slurm.conf.html#lbAJ:
> NOTE: Standard output and error messages are normally not preserved. Explicitly write output and error messages to an appropriate location if you wish to preserve that information.

DriesVerachtert: Yeah on BB5 we also had to do such tricks :-)

launch_system/cloud-init.cfg
```diff
@@ -29,8 +29,111 @@ runcmd:
   - grep -q "${fsx_dns_name}" /etc/fstab || echo "${fsx_dns_name}@tcp:/${fsx_mount_name} /data/scratch/ lustre defaults,_netdev 0 0" >> /etc/fstab
   - mount /data/scratch
   - mkdir -p /data/aws_s3_open
-  - mkdir -p /data/aws_s3_internal
+  - mkdir -p /data/aws_s3_internal/public
   - mount -t efs -o tls,accesspoint=${opendata_access_point_id} ${publicdata_efs_id}:/ /data/aws_s3_open
-  - mount -t efs -o tls,accesspoint=${publicdata_access_point_id} ${publicdata_efs_id}:/ /data/aws_s3_internal
+  - mount -t efs -o tls,accesspoint=${publicdata_access_point_id} ${publicdata_efs_id}:/ /data/aws_s3_internal/public
```
GianlucaFicarelli, line 36: This change is intended to mount /public and /private separately, but it needs to be verified if the public data is mounted correctly.
In the sandbox, amazon-efs-utils is missing because it's not installed in the AMI and not listed in the packages in cloud-init.cfg so the EFS mount fails.

PR #1580, opened 2026-07-09
Allow Launch System API to reach accounting service, reject non-allowlisted IPs with 403

pgetta:
## What

- Export the launch_system API (trusted) subnet CIDRs as `api_subnet_cidr_blocks` and add them to the accounting service ALB `source_ip` allowlist, so the Launch System API can call the accounting service through the private ALB.
- Add an explicit `fixed-response` deny rule for `/api/accounting*` at priority 690: requests from IPs outside the allowlist now get `403 {"detail": "Forbidden"}` instead of falling through to the listener default action.

## Notes

- The allowlist grows from ~12 to ~14 CIDRs, so the chunked forward rules gain a 4th rule at priority 653 (chunks of 4, base 650). The deny rule sits at a fixed offset (base + 40 = 690) so allowlist growth can never collide with it, and stays below the next service's priority (700, cells_svc).
- 403 chosen over 404/401: the request is refused on authorization grounds (IP allowlist), matching AWS WAF convention and keeping allowlist misconfigurations easy to diagnose on the internal ALB.

P.S. It turns out that [Only allow a list of services to access accounting and launch-system
](https://github.com/openbraininstitute/aws-terraform-deployment/pull/1281) had introduced a regression where the Launch System API could no longer communicate with the accounting service which caused an issue with reservation cancellation for failed tasks.

### DIFF
--- a/accounting_svc/alb.tf
+++ b/accounting_svc/alb.tf
@@ -43,3 +43,28 @@ resource "aws_lb_listener_rule" "accounting_private_listener_rule" {
     }
   }
 }
+
+# Requests to the accounting path that did not match any source_ip allowlist
+# rule above are explicitly rejected instead of falling through to the
+# listener default action. Priority must stay above the highest possible
+# chunk rule (base + chunk_index) and below other services' rules (700+).
+resource "aws_lb_listener_rule" "accounting_private_listener_rule_deny" {
+  listener_arn = var.private_alb_listener_arn
+  priority     = var.listener_rule_base_priority + 40
+
+  action {
+    type = "fixed-response"
+
+    fixed_response {
+      content_type = "application/json"
+      message_body = "{\"detail\": \"Forbidden\"}"
+      status_code  = "403"
+    }
+  }
+
+  condition {
+    path_pattern {
+      values = ["${var.root_path}*"]
+    }
+  }
+}
--- a/launch_system/outputs.tf
+++ b/launch_system/outputs.tf
@@ -26,6 +26,14 @@ output "orchestrator_subnet_cidr_blocks" {
   ]
 }
 
+output "api_subnet_cidr_blocks" {
+  description = "CIDR blocks of the launch_system API (trusted) subnets"
+  value = [
+    aws_subnet.trusted_a.cidr_block,
+    aws_subnet.trusted_b.cidr_block,
+  ]
+}
+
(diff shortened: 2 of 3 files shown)


pgetta:
Relates to https://github.com/openbraininstitute/prod-virtual-lab/issues/361.

accounting_svc/alb.tf
```diff
@@ -43,3 +43,28 @@ resource "aws_lb_listener_rule" "accounting_private_listener_rule" {
     }
   }
 }
+
+# Requests to the accounting path that did not match any source_ip allowlist
+# rule above are explicitly rejected instead of falling through to the
+# listener default action. Priority must stay above the highest possible
+# chunk rule (base + chunk_index) and below other services' rules (700+).
+resource "aws_lb_listener_rule" "accounting_private_listener_rule_deny" {
```
danifr, line 51: Do we need this explicit deny? Without it, it'll just fall through to the default action which is a 302 redirect to the main domain hostname no?

I am asking because we don't have in any other services.

pgetta: So that default redirect is what had hidden the issue for quite some time. Because the Launch System API was getting 302 -> 200 thus was not reporting any issues while the requests were never actually reaching accounting API.

My feeling is that we should have this for every similar service. I can note this down and refactor the rest some time later if that's fine with you.

pgetta: Ok, I'm not so sure now if it was getting status 200 at the end, but clear 4xx or 5xx would clearly signal to the accounting that something is wrong.

pgetta: And probably it's not the best idea to let the requests intended to reach the Accounting service silently land to the Core Web app.

pgetta: Of course an additional rule will have an impact on the ALB costs, but I hope it's negligible.

danifr: OK yeah I think it makes sense indeed. From what I am seeing can add one deny to launch_system/alb.tf, but if you want we can do it in a different PR

danifr (approved):

accounting_svc/alb.tf
pgetta: I'll create a separate PR

pgetta: Thanks for the approval!

